PR #9, opened 2020-04-06
Feature/event props

Christos-Zacharopoulos:
This PR adds a generic EventProps type for onClick and onBlur events and adds them to Button component.

### DIFF
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -1,2 +1,8 @@
 /** A type to turn any type optional properties to required */
 export type RequiredProperties<T> = { [K in keyof T]-?: T[K] };
+
+/** A generic type for native events */
+export type EventProps = {
+  onClick?: () => void;
+  onBlur?: () => void;
+};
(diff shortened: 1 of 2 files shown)


src/utils/common.ts
```diff
@@ -1,2 +1,8 @@
 /** A type to turn any type optional properties to required */
 export type RequiredProperties<T> = { [K in keyof T]-?: T[K] };
+
+/** A generic type for native events */
+export type EventProps = {
+  onClick?: () => void;
```
panvourtsis, line 6: Can we pass the event or no ?

panvourtsis: on both events i mean

Christos-Zacharopoulos: Yes we can. But then it will be specific to button since the event is 
` React.MouseEvent<HTMLButtonElement, MouseEvent>`
instead of simply 
`React.MouseEvent`

PR #12, opened 2020-04-09
Feature/pagination

### LINKED ISSUE #6: Pagination
Introduce a component to handle pagination of any list or table data.

This will be based on the zeplin design of any screen
image

An example can be found [here](https://zpl.io/VxeGvMJ)

panvourtsis:
In this PR i am introducing the pagination as a component and also closes #6 . 

The basic idea is to have a component that returns the page on a callback and controls the pages internally. I also had to touch some colors and the `disabled` state of buttons.

### DIFF
--- a/src/hooks/usePagination.ts
+++ b/src/hooks/usePagination.ts
@@ -1,0 +1,63 @@
+import { useCallback, useState } from 'react';
+
+type Props = {
+  page: number;
+  count: number;
+  onChange: (page: number) => void;
+};
+
+const usePagination = ({ page = 1, count = 1, onChange: handleChange }: Props) => {
+  const [currentPage, setCurrentPage] = useState(page);
+  const nextPage = currentPage + 1;
+  const prevPage = currentPage - 1;
+  const hasNextPage = nextPage <= count;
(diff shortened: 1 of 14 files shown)


src/hooks/usePagination.ts
```diff
@@ -0,0 +1,63 @@
+import { useCallback, useState } from 'react';
+
+type Props = {
+  page: number;
+  count: number;
+  onChange: (page: number) => void;
+};
+
+const usePagination = ({ page = 1, count = 1, onChange: handleChange }: Props) => {
```
Christos-Zacharopoulos, line 9: usePagination hook doesn't update properly the value to the outer world. Either changes to the dependency arrays or a useEffect to trigger the onChange function.

panvourtsis: nicely done catching this, just fixed with a cleaner solution by tracking the value change and trigger handleChange there

mkarajohn:
Is the pagination supposed to be based on url parameters?

mkarajohn:
Also, tbh, I don't see how a pagination component is part of a design system. I have objections regarding the matter

PR #16, opened 2020-04-13
Feature/table enhanced

panvourtsis:
This PR will introduce several things required for all projects to come.

- Import the new Checkbox as the checkbox of the table for selecting rows
- On row expand use all the available space under the row for showing content
- Return the row state for selected and expanded on expanded function callback

### DIFF
--- a/src/components/CheckBox/CheckBox.tsx
+++ b/src/components/CheckBox/CheckBox.tsx
@@ -34,7 +34,9 @@ const CheckBox: React.FC<Props> = ({
   const handleInputChange = (event: ChangeEvent) => {
     const newChecked = !isChecked;
 
-    setIsChecked(newChecked);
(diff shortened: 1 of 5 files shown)


src/components/CheckBox/CheckBox.tsx
```diff
@@ -50,7 +52,7 @@ const CheckBox: React.FC<Props> = ({
           type="checkbox"
           onChange={handleInputChange}
           disabled={disabled}
-          checked={isChecked}
+          checked={checked}
```
Christos-Zacharopoulos, line 55: Since we are using checked prop, I think we don't need the useState anymore.

panvourtsis: it doesnt hurt to have controlled and uncontrolled version

PR #24, opened 2020-04-29
Table Row Align Cell prop

Christos-Zacharopoulos:
This PR adds the align prop to Cell in order to customize the alignment  of the Cell.

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -17,6 +17,7 @@ export type Cell<T> = {
   content: number | string | ContentComponent<T>;
   colSpan?: number;
   type?: 'financial' | 'normal';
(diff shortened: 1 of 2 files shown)


src/components/Table/Table.tsx
```diff
@@ -17,6 +17,7 @@ export type Cell<T> = {
   content: number | string | ContentComponent<T>;
   colSpan?: number;
   type?: 'financial' | 'normal';
+  align?: 'left' | 'right';
```
panvourtsis, line 20: as this is align the proper way is to have a 'center' option as well

Christos-Zacharopoulos: I know but it is not a value we use in the table cell component before this addition and I don't want to add any more to this.

PR #32, opened 2020-05-22
add optional type on input

panvourtsis:
This PR adds type to all input to allow to change text to password etc


Christos-Zacharopoulos:
Just one comment. I thought we had discussed and decided to not destructure props inside components.

panvourtsis:
did we? i dont have a problem but maybe as this is an external project such things must be defined to the linting options

PR #34, opened 2020-06-03
Feature/popover menu

panvourtsis:
This PR introduce the popover menu component. 

Here we have a button that will be either a normal one or a button with an icon at the right. Such case only shows in VH for now but it will be used in all project thus is being included here.

I added a new property to button component for icon position and the new vertical dots icon for this scenario. Now the button can handle icon on the left or right of the text.

image


src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+                color={buttonType === ('primary' || 'secondary') ? 'dark' : 'light'}
+              />
+            ) : null
+          }
+        >
+          <span>{buttonText}</span>
+        </Button>
+        {open && (
+          <div
+            css={css`
```
Christos-Zacharopoulos, line 66: I am not a big fan of having css here like this. It's not a deal breaker but I just wanted to point it out.  :nerd_face:

src/components/Icon/assets/iconSelector.ts
```diff
@@ -61,6 +62,7 @@ const iconSelector = {
   checkedList: CheckedListIcon,
   hourglass: HourglassIcon,
   ticTacToeArrow: TicTacToeArrowIcon,
```
Christos-Zacharopoulos, line 64: At some point we need to ask the UX team for proper names on the icons.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,116 @@
...
+              width: ${rem(148)};
+              height: auto;
+              background-color: #fff;
+              box-shadow: 0px 0px ${rem(16)} grey;
+              display: flex;
+              flex-direction: column;
+              justify-content: space-evenly;
+              text-align: center;
+              border-radius: ${rem(4)};
+              z-index: 1;
```
Christos-Zacharopoulos, line 80: Maybe this should be initialized from the theme so it can be configured according to each project's needs and special cases.

src/components/Icon/assets/iconSelector.ts
panvourtsis: i dont think its their job exactly but we need to have the same names for reference

src/components/Menu/Menu.tsx
panvourtsis: indexing ?

src/components/Menu/Menu.tsx
panvourtsis: haters gonna hate - we can move it to the css file if needed but according to our last team discussion we said that this is fine and for me it looks ok tbh

src/components/Menu/Menu.tsx
Christos-Zacharopoulos: yes

panvourtsis:
maybe add specific height for menu if the items are many?

src/components/Menu/Menu.tsx
panvourtsis: you could make the parent higher :) rather than changing the component

Christos-Zacharopoulos: But you would have to add an extra layer for this. 

I like the way material allows full customization over all of the components they have. 

If we are to leave it like this and not have some kind of manipulating/changing some values we should ask for the hierarchy of the components according to the importance they have and style them accordingly. So at least, when they are on the same level, they will display properly with no overlaps. 

(This is not something to be implemented on this specific PR. Maybe discuss this further on an issue?)

panvourtsis: i see the importance in this and i created a new issue for it on #35 for further discussion and you can put your opinion or thoughts or even documentation in there

src/components/Button/Button.style.ts
```diff
@@ -59,14 +65,23 @@ export const buttonStyle = ({
...
 });
+
+export const iconStyle = ({
+  size,
+  iconAlign,
+}: RequiredProperties<Props & { hasChildren: boolean }>) => (theme: Theme) => {
+  const margin = size === 'sm' ? theme.spacing.sm : theme.spacing.md;
+
+  return {
+    marginLeft: iconAlign === 'left' ? margin : theme.spacing.sm,
```
michalisktr, line 84: You could refactor this line of code into an arrow function in the iconStyle's scope since the line below is almost identical.

example:
```
const getMargin = (alignPosition) => iconAlign === alignPosition ? margin : theme.spacing.sm;
....
return {
marginLeft: getMargin('left'),
marginRight: getMargin('right')
}
```

Christos-Zacharopoulos:
Add custom scrollbars maybe. Some of us don't use macs and have to witness this: 

scroll

src/components/Menu/Menu.style.ts
```diff
@@ -0,0 +1,72 @@
+import { backgroundPickerBasedOnType, colorPickerBasedOnType } from 'utils/themeFunctions';
+import { Props } from 'components/Button/Button';
+import { RequiredProperties } from 'utils/common';
+import { Theme } from 'theme';
+import { FlexDirectionProperty } from 'csstype';
+import { rem } from 'polished';
+
+/** Calculates the button specific height based on the size passed to it
+ * These sizes are specific to this button thus these are placed here and not in the config **/
+const heightBasedOnSize = (size: 'lg' | 'md' | 'sm') => {
```
michalisktr, line 10: Since this function implementation is repeated in components Button, IconButton and Menu you could move this into a helper function. 
Furthermore, you could refactor "switch" with bracket notation which, i think, it is more efficient.   

example in common.ts:
```
export interface SizeConfig {
  lg: string;
  md: string;
  sm: string;
}
```

example in helpers.ts 
```
const SIZE_OPTIONS: SizeConfig = {
  lg: rem(56),
  md: rem(46),
  sm: rem(40),
};

export const heightBasedOnSize = (size: 'lg' | 'md' | 'sm'): string => SIZE_OPTIONS[size];
```

src/components/Menu/Menu.style.ts
```diff
@@ -0,0 +1,72 @@
...
+  filled,
+  size,
+  icon,
+  disabled,
+  childrenCount,
+}: RequiredProperties<Props & { childrenCount: number }>) => (theme: Theme) => {
+  const calculatedPaddingSpace = size === 'sm' ? theme.spacing.md : theme.spacing.lg;
+  const calculatedPaddingSpaceIfIcon = size === 'sm' ? theme.spacing.sm : theme.spacing.md;
+
+  const defineBackgroundColor = (): string => {
```
michalisktr, line 32: You could move this function (_defineBackgroundColor_) into the helpers.ts.

Issue #5, opened 2020-03-26
Breadcrumb component


panvourtsis:
This will be a component in order to show the navigation to the user. It will be consisted by 3 types of breadcrumb, simple, simple with options and advanced

All dropdowns, spacings and buttons that are going to used are going to be from the design system.

See reference below:
![image](https://user-images.githubusercontent.com/6433679/84120161-18439d00-aa3e-11ea-81e8-85de8b0cd00b.png)

elektracodes:
Calling dibs on that one

michalisKout:
![](https://github.trello.services/images/mini-trello-icon.png) [Issue #5 https://github.com/Orfium/orfium-ictinus/issues/5](https://trello.com/c/NQKm1Gc2/10-issue-5-https-githubcom-orfium-orfium-ictinus-issues-5)

panvourtsis:
do we have an official board for this project?

PR #36, opened 2020-06-10
Feature/breadcrumb

michalisKout:
this is an implementation of the issue #5

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.stories.mdx
+++ b/src/components/Breadcrumb/Breadcrumb.stories.mdx
@@ -48,6 +48,8 @@ A simple Breadcrumb example
         <Breadcrump>
           <div>First Level</div>
           <div>Second Level</div>
+          <div>Third Level</div>
+          <div>Forth Level</div>
           <div>Current Level</div>
         </Breadcrump>
     </Stack>
--- a/src/components/Breadcrumb/Breadcrumb.stories.mdx
+++ b/src/components/Breadcrumb/Breadcrumb.stories.mdx
@@ -1,6 +1,6 @@
 import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
 import { withKnobs, boolean, array, select, text } from "@storybook/addon-knobs";
-import Breadcrump from './Breadcrumb';
+import Breadcrumb from './Breadcrumb';
 import Stack from '../storyUtils/Stack';
 import {
   BrowserRouter as Router,
@@ -37,7 +37,7 @@ import { Breadcrumb } from 'orfium-design-system';
 
 # Props
 
-<Props of={() => true} />
+<Props of={Breadcrumb} />
 
 
 # Simple Breadcrumbs (up to 3 levels)
@@ -46,13 +46,13 @@ A simple Breadcrumb example
 <Preview>
   <Story name="Simple Breadcrumbs">
     <Stack>
-        <Breadcrump>
+        <Breadcrumb>
           <div>First Level</div>
           <div>Second Level</div>
           <div>Third Level</div>
           <div>Forth Level</div>
           <div>Current Level</div>
-        </Breadcrump>
+        </Breadcrumb>
     </Stack>
   </Story>
 </Preview>
@@ -64,7 +64,7 @@ A simple Breadcrumb with n levels example
   <Story name="Simple Breadcrumbs with options">
     <Stack>
       <Router>
-          <Breadcrump data={[
+          <Breadcrumb data={[
             { to: '/', label: 'Home' },
             { to: '/first-level', label: 'First Level' },
             { to: '/second-level', label: 'Second Level' },
--- a/src/components/Breadcrumb/Breadcrumb.stories.mdx
+++ b/src/components/Breadcrumb/Breadcrumb.stories.mdx
@@ -3,6 +3,7 @@ import { withKnobs, boolean, array, select, text } from '@storybook/addon-knobs'
 import Breadcrumb from './Breadcrumb';
 import Stack from '../storyUtils/Stack';
 import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
+import BreadcrumbShowcase from '../storyUtils/BreadcrumbShowcase/BreadcrumbShowcase.tsx';
 
 <Meta title="Design System|Breadcrumb" component={() => true} />
 
@@ -109,6 +110,31 @@ A advanced Breadcrumb with dd levels example
           <Route path="/forth-level">{() => <div>forth level</div>}</Route>
         </Switch>
       </Router>
+    </Stack>
+  </Story>
+</Preview>
+
+# Breadcrumb showcase (with dd in levels)
+
+A Breadcrumb showcase with dd levels example
+
+<Preview>
+  <Story name="Breadcrumb showcase">
+    <Stack>
+      <BreadcrumbShowcase
+        initData={[
+          { to: '/first-level', label: 'First Level', options: ['ITEM 1', 'ITEM 2'] },
+          { to: '/second-level', label: 'Second Level', options: ['ITEM 1', 'ITEM 2'] },
+          { to: '/third-level', label: 'Third Level', options: ['ITEM 1', 'ITEM 2'] },
+          { to: '/forth-level', label: 'Forth Level', options: ['ITEM 1', 'ITEM 2'] },
+          {
+            to: '/fifth-level',
+            label: 'Fifth Level',
+            options: ['ITEM 1', 'ITEM 2'],
+            onChangeHandler: text => text && alert(text),
+          },
+        ]}
+      />
     </Stack>
   </Story>
 </Preview>
--- a/src/components/Breadcrumb/Breadcrumb.stories.mdx
+++ b/src/components/Breadcrumb/Breadcrumb.stories.mdx
@@ -50,8 +50,6 @@ A simple Breadcrumb example
           <div>First Level</div>
           <div>Second Level</div>
           <div>Third Level</div>
-          <div>Forth Level</div>
-          <div>Current Level</div>
         </Breadcrumb>
     </Stack>
   </Story>
--- a/src/components/Breadcrumb/Breadcrumb.stories.mdx
+++ b/src/components/Breadcrumb/Breadcrumb.stories.mdx
@@ -63,7 +63,7 @@ A simple Breadcrumb with n levels example
     <Stack>
       <Router>
           <Breadcrumb data={[
-            { to: '/first-level', label: 'First Level'},
+            { to: '/first-level', label: 'First Level' },
             { to: '/second-level', label: 'Second Level' },
             { to: '/third-level', label: 'Third Level' },
             { to: '/forth-level', label: 'Forth Level' },
@@ -96,9 +96,31 @@ A simple Breadcrumb with n levels example
 A advanced Breadcrumb with dd levels example
 
 <Preview>
-  <Story name="Advanced Breadcrumbs">
+  <Story name="Advanced Breadcrumbs with options per data item">
     <Stack>
-        <div>Breadcrumb placeholer</div>
+      <Router>
+          <Breadcrumb data={[
+            { to: '/first-level', label: 'First Level', options: ['ITEM 1', 'ITEM 2']},
+            { to: '/second-level', label: 'Second Level', options: ['ITEM 1', 'ITEM 2'] },
+            { to: '/third-level', label: 'Third Level', options: ['ITEM 1', 'ITEM 2'] },
+            { to: '/forth-level', label: 'Forth Level', options: ['ITEM 1', 'ITEM 2'] },
+            { to: '', label: 'Fifth Level', options: ['ITEM 1', 'ITEM 2'], changeHandler: (text) => text && alert(text) },
+          ]}/>
+          <Switch>
+                <Route path="/first-level">
+                  {() => <div>first level</div>}
+                </Route>
+                <Route path="/second-level">
+                  {() => <div>second level</div>}
+                </Route>
+                <Route path="/third-level">
+                  {() => <div>third level</div>}
+                </Route>
+                <Route path="/forth-level">
+                  {() => <div>forth level</div>}
+                </Route>
+          </Switch>
+      </Router>
     </Stack>
   </Story>
 </Preview>
--- a/src/components/Breadcrumb/Breadcrumb.stories.mdx
+++ b/src/components/Breadcrumb/Breadcrumb.stories.mdx
@@ -14,7 +14,7 @@ Description of Breadcrumb component
 ```js
 import { Breadcrumb } from 'orfium-design-system';
 
-<div>Breadcrumb placeholer</div>
+<Breadcrumb><div>content</div></Breadcrumb>
 ```
 
 # Props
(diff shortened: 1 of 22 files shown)


src/components/Breadcrumb/Breadcrumb.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+  BrowserRouter as Router,
+  Switch,
+  Route,
+} from "react-router-dom";
+
+<Meta title="Design System|Breadcrumb" component={() => true} />
+
+# Breadcrumb
+
+Description of Breadcrumb component
```
panvourtsis, line 15: introduce some description here

src/components/Breadcrumb/Breadcrumb.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+```js
+import { Breadcrumb } from 'orfium-design-system';
+
+<Breadcrumb data=[{to:"/path", label: "label"}]/>
+```
+
+# Props
+
+<Props of={() => true} />
```
panvourtsis, line 39: import the props of the component in order to show on the showcase

src/components/Breadcrumb/Breadcrumb.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+                </Route>
+          </Switch>
+      </Router>
+    </Stack>
+  </Story>
+ 
+</Preview>
+
+# Advanced Breadcrumbs (with dd in levels)
+A advanced Breadcrumb with dd levels example
```
panvourtsis, line 94: is this finished ?

src/components/Breadcrumb/Breadcrumb.stories.mdx
```diff
@@ -0,0 +1,102 @@
...
+<Breadcrumb data=[{to:"/path", label: "label"}]/>
+```
+
+# Props
+
+<Props of={() => true} />
+
+
+# Simple Breadcrumbs (up to 3 levels)
+A simple Breadcrumb example
```
panvourtsis, line 43: this case must be up to 3 levels

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,74 @@
...
+import useTheme from 'hooks/useTheme';
+import BreadcrumbItem from './BreadcrumbItem/BreadcrumbItem';
+import isEmpty from 'lodash/isEmpty';
+import uniqueId from 'lodash/uniqueId';
+import { Link } from 'react-router-dom';
+import BreadcrumbCollapsed from './BreadcrumbCollapsed/BreadcrumbCollapsed';
+
+export type Props = {
+  separatorContent?: '*' | '>' | '/';
+  data: [];
```
panvourtsis, line 14: define data so you can use it as `BreadcrumbItemData[]`

e.g 
```
type BreadcrumbItemData = {
  to: string;
  label: string;
}
```

OPTIONAL: data literal not needed as these types are models so `BreadcrumbItem` should do the trick

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,74 @@
...
+  breadcrumbItemClickHandler: () => void;
```
panvourtsis, line 15: what is this ?

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,74 @@
...
+import React, { useMemo } from 'react';
+import { breadcrumbStyles, breadcrumbLinkStyles } from './Breadcrumb.style';
+import useTheme from 'hooks/useTheme';
+import BreadcrumbItem from './BreadcrumbItem/BreadcrumbItem';
+import isEmpty from 'lodash/isEmpty';
+import uniqueId from 'lodash/uniqueId';
+import { Link } from 'react-router-dom';
+import BreadcrumbCollapsed from './BreadcrumbCollapsed/BreadcrumbCollapsed';
+
+export type Props = {
```
panvourtsis, line 12: add some description per prop as it helps on the documentation

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,74 @@
...
+  );
+
+  const childrenCollection = React.Children.toArray(children);
+  const dataItems = isEmpty(data) ? childrenCollection : data.map(passDataToRouterLink);
+
+  const isLastItem = (itemIndex: number) => itemIndex === dataItems.length - 1;
+  const shouldCollapse = (item: BreadcrumbItem, itemIndex: number) =>
+    item && dataItems.length > 4 && itemIndex > 0 && itemIndex < dataItems.length - 2;
+
+  const collapsedItems = useMemo(() => dataItems.filter(shouldCollapse), []);
```
panvourtsis, line 41: i think this should be updated when data is

src/components/Breadcrumb/BreadcrumbCollapsed/BreadcrumbCollapsed.style.ts
```diff
@@ -0,0 +1,31 @@
...
+  &:hover {
+    transition: 0.5s all;
+    background: ${theme.palette.gray50};
+  }
+  height: ${rem(32)};
+  font-size: ${theme.typography.fontSizes['14']};
+  list-style-type: none;
+  padding-top: ${rem(16)};
+  height: ${rem(32)};
+  font-size: 0.875rem;
```
panvourtsis, line 30: use our typography here

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { breadcrumbStyles, breadcrumbLinkStyles } from './Breadcrumb.style';
+import useTheme from 'hooks/useTheme';
+import BreadcrumbItem from './BreadcrumbItem/BreadcrumbItem';
+import isEmpty from 'lodash/isEmpty';
+import uniqueId from 'lodash/uniqueId';
+import { Link } from 'react-router-dom';
+import BreadcrumbCollapsed from './BreadcrumbCollapsed/BreadcrumbCollapsed';
+
+export type Props = {
+  separatorContent?: '*' | '>' | '/';
```
panvourtsis, line 13: as this is used elsewhere define it as type and use it on your component

src/components/Breadcrumb/BreadcrumbItem/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { BreadcrumbItemData } from '../Breadcrumb';
+
+type Props<T> = {
+  childComponent: T;
+  isLastItem: boolean;
+  separatorContent: '*' | '>' | '/';
+  clickHandler?: () => void;
+};
+
+const BreadcrumbItem: React.FC<Props<BreadcrumbItemData | React.ReactNode>> = props => {
```
panvourtsis, line 16: the generic in Props i think is not right? ReactNode and BreadcrumbItemData are not the same or event partially equal

src/components/Breadcrumb/Separator/Separator.style.ts
```diff
@@ -0,0 +1,10 @@
+import { Theme } from 'theme';
+import { css } from '@emotion/core';
+
+// import { RequiredProperties } from '../../utils/common';
```
panvourtsis, line 4: you can remove unused code

src/components/Breadcrumb/Breadcrumb.stories.mdx
```diff
@@ -0,0 +1,102 @@
+import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
```
panvourtsis, line 1: also if i can add those two screenshots that are not the same as the design provided

Screenshot 2020-06-15 at 1 29 13 PM

design provided
image

Screenshot 2020-06-15 at 1 29 10 PM

src/components/Breadcrumb/Breadcrumb.stories.mdx
michalisKout: This is not implemented, I would like to further discuss the approach of the implementation.

src/components/Breadcrumb/Breadcrumb.stories.mdx
michalisKout: I will fix the margins, the width, the text alignment and the shadow. Are there any colour codes for the collapsed view and the hovered areas? Should we use a svg instead of text (3 dots) for the collapsed items ?

src/components/Breadcrumb/Breadcrumb.tsx
michalisKout: Initially was a clickHandler for a breadcrumb but as the implementation of the breadcrumb changed this became an obsolete feature. I will just remove it.

src/components/Breadcrumb/BreadcrumbItem/BreadcrumbItem.tsx
michalisKout: The original purpose was to pas a generic type to the breadcrumbItem component in order to make the implementation more flexible to other types of data. As you correctly notice, this is not necessary anymore as childComponent will always be a React.ReactNode type.

Besides the above, I also think that I implemented it wrong... The main idea was something like this:
```
type Props<T extends React.ReactNode> = {
  childComponent: T;
  isLastItem: boolean;
  separatorContent: SeparatorStyle;
};

const BreadcrumbItem = <T extends React.ReactNode>(props: Props<T>) => ...
```

src/utils/helpers.ts
```diff
@@ -13,3 +15,7 @@ export function isComponentFunctionType(
 ): element is Function {
   return typeof element === 'function' && React.isValidElement(element());
 }
+
+export const getLastDataItemProperty = (data: BreadcrumbItemData[], propertyPath: string) => {
```
panvourtsis, line 19: a small problem here is that this function is listed under the global helpers. Thus anyone that will try to use it will have to "feed it" that exact type of data. I feel that this is more of a local helper for breadcrumb.

Also this can be avoided by using lodash pick or omit 
```_.pick(_.last(data), ['label', 'onChangeHandler', 'options'])```

File: src/components/Breadcrumb/Breadcrumb.tsx [hunk 15]
```diff
@@ -0,0 +1,77 @@
...
+  const dataItems = isEmpty(data) ? childrenCollection : data.map(passDataToRouterLink);
+
+  const isLastItem = (dataItems: React.ReactNode[], itemIndex: number) =>
+    itemIndex === dataItems.length - 1;
+  const shouldCollapse = (item: React.ReactNode, itemIndex: number) =>
+    item && dataItems.length > 4 && itemIndex > 0 && itemIndex < dataItems.length - 2;
+
+  const collapsedItems = useMemo(() => dataItems.filter(shouldCollapse), [dataItems]);
+  const lastItem = {
+    label: getLastDataItemProperty(data, 'label'),
```
panvourtsis, line 37: see my comment on the definition of that function to pick the last item - an alternative solution

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,77 @@
...
+    <Link css={breadcrumbLinkStyles()(theme)} key={to} to={to}>
+      {label}
+    </Link>
+  );
+  const childrenCollection = React.Children.toArray(children);
+  const dataItems = isEmpty(data) ? childrenCollection : data.map(passDataToRouterLink);
+
+  const isLastItem = (dataItems: React.ReactNode[], itemIndex: number) =>
+    itemIndex === dataItems.length - 1;
+  const shouldCollapse = (item: React.ReactNode, itemIndex: number) =>
```
panvourtsis, line 32: please add a small description above this to explain the logic 
e.g Checks if items are between 1-3 and if should break the breadcrumb items

just for future reference in 3-4 months this would be helpful

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export type Props = {
+  /** Defines the data for constructing the related breadcrumb items */
+  data: BreadcrumbItemData[];
+};
+
+const Breadcrumb: React.FC<Props> = props => {
+  const { children, data = [] } = props;
+  const theme = useTheme();
+  const passDataToRouterLink = ({ to, label }: BreadcrumbItemData) => (
```
panvourtsis, line 22: you can move this outside the breadcrumb as you will need it only once

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <BreadcrumbCollapsed collapsedItems={collapsedItems} key={itemKey} />
+        ) : null;
+      }
+
+      const isLast = isLastItem(dataItems, index);
+
+      return isLast ? (
+        <BreadcrumbItem
+          key={itemKey}
+          lastItemLabel={lastItem.label}
```
panvourtsis, line 57: OPTIONAL - it could be simpler if you wanted to merge 
lastItemLabel and isLastItem if you just pass down the last item, if no last item then you know

src/components/Breadcrumb/BreadcrumbItem/BreadcrumbItem.style.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    & > span {
+      margin-left: 0;
+    }
+  }
+
+  & > div > div > div {
+    top: 3rem;
+    box-shadow: #8080802b 0px 0px ${rem(16)};
```
panvourtsis, line 33: we must define these colors as shadow colors - there is the same in another file as well

src/components/Menu/Menu.style.ts
```diff
@@ -70,3 +71,36 @@ export const buttonSpanStyle = ({
     marginRight: hasChildren ? theme.spacing.sm : 0,
   },
 });
+
+type MenuOptions = {
+  menuPosition?: 'left' | 'right';
```
panvourtsis, line 76: add a type in the menu.tsx that would be like the below and then use it everywhere you want

```
type MenuPositionAllowed = 'left' | 'right';

type Props = {
  ...
  menuPosition?: MenuPositionAllowed;
  ...
}
```

src/utils/helpers.ts
michalisKout: I agree this is more like a local helper. Should i create a util.ts file inside the breadcrumb directory ? 

Great idea on refactoring it this way.

src/components/Breadcrumb/Breadcrumb.tsx
michalisKout: if I move this outside of the breadcrumb, I will have to convert this util function to a functional component. Therefore, I will have to create another function to use it for the data mapping. 

example: 
```
data.map( ( {label, to}: BreadcrumbItemData) => <PassDataToRouteLink label={label} to={to} /> );
```
We could try wrapping the passDataToRouterLink with a useCallback hook but i am still not sure if we gain performance value.

src/components/Breadcrumb/Breadcrumb.tsx
michalisKout: Yep, i made it simpler, you can check it out

src/components/Menu/Menu.style.ts
panvourtsis: this must be imported from the menu.tsx

PR #43, opened 2020-07-21
wrap inline header content to span

panvourtsis:
This PR wraps the content of the inline header table cell to a span in order for the QA to be able to "catch" the contents with selenium and xPath

### DIFF
--- a/src/components/Table/components/RenderRowOrNestedRow/RenderRowOrNestedRow.tsx
+++ b/src/components/Table/components/RenderRowOrNestedRow/RenderRowOrNestedRow.tsx
@@ -75,7 +75,7 @@ const RenderRowWithCells = React.memo(({ checked = false, toggleChecked = () => 
             {isComponentFunctionType(content) ? (
               content({ content, colSpan })
             ) : (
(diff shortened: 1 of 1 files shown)


src/components/Table/components/RenderRowOrNestedRow/RenderRowOrNestedRow.tsx
```diff
@@ -72,7 +72,11 @@ const RenderRowWithCells = React.memo(({ checked = false, toggleChecked = () =>
                 {columns[cellCounter]}
               </div>
             )}
-            {isComponentFunctionType(content) ? content({ content, colSpan }) : content}
+            {isComponentFunctionType(content) ? (
+              content({ content, colSpan })
+            ) : (
+              <span>{content}</span>
```
Christos-Zacharopoulos, line 78: Wouldn't it be better to pass an id to this?

panvourtsis: as i discussed with the QA data test id only needed for the parent wrappers

PR #51, opened 2020-09-01
add keyword icon

panvourtsis:
This PR simply adds a new icon


src/components/Icon/assets/keyword.svg
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <g id="🔶-Symbols" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Keyword" fill="#232323">
```
Christos-Zacharopoulos, line 4: I am not sure but I think this fill needs to be removed so that that svg can be styled with different colors

Christos-Zacharopoulos (approved):

src/components/Icon/assets/keyword.svg
panvourtsis: Screenshot 2020-09-01 at 2 56 13 PM
Screenshot 2020-09-01 at 2 56 26 PM
Screenshot 2020-09-01 at 2 56 43 PM

PR #53, opened 2020-09-02
Feature/range picker

### LINKED ISSUE #3: Date / Time pickers
WIP

panvourtsis:
This PR introduce the new Date Range Picker / Day Picker

Time is not yet implemented due to the fact that we miss the dropdown selection for numeric values etc.

Closes #3 

Design can be found [here](https://share.goabstract.com/43803ace-7fbd-4baf-a5f9-5dee5acfcfe7?collectionLayerId=d530c127-76af-4b97-8160-557cff67a7a8&mode=design)

### Implementation
For the implementation of the date picker and for future use of the time and date i added _[dayjs](https://www.npmjs.com/package/dayjs)_ as a simple lightweight lib in order to manipulate dates and time.

I used the _[react-day-picker](https://react-day-picker.js.org/examples/selected-range/)_ as the base of the component in order to avoid starting something from the scratch. This way we can have a keyboard friendly already tested solution for our system. The **bad thing about it** is that in order to style it you must use _CSS_. If you see something you don't like about it let me know.

### TextField 
The TextField as shown from the design is the new outlined. This is not yet available thus we show the original input that we have

### New Config
Due to the fact we use some CSS not for this i changed the storybook config slightly for the SVGs in order to be able to have them as components in the _.tsx_ files but to also import them as images with _url-loader_ on _style.ts_ files.

### New Spacing
I added a new variation to spacings as we already discussed. We use to have spacings like
* 8
* 16
* 32

Now we introduce a new value - 24 ! Good times.
In order to update that i had to push everything a bit and change some files that were using the old spacing. 
This will be updated as a new PR with the palette in next versions


src/components/DatePicker/DatePicker.style.ts
```diff
@@ -0,0 +1,99 @@
+import { Theme } from 'theme';
+import { css, SerializedStyles } from '@emotion/core';
+import FatArrowLeft from '../Icon/assets/fat-arrow-left.svg';
+import FatArrowRight from '../Icon/assets/fat-arrow-right.svg';
+
+export const optionStyle = ({ selected }: { selected?: boolean }) => (
+  theme: Theme
```
michalisKout, line 7: you can remove the unused variable `theme`

src/components/DatePicker/DatePicker.style.ts
```diff
@@ -0,0 +1,99 @@
...
+): SerializedStyles => css`
+  white-space: nowrap;
+  padding: 15px;
```
michalisKout, line 10: You could use the theme.spacing here instead.

src/components/DatePicker/OverlayComponent/OverlayComponent.tsx
```diff
@@ -0,0 +1,55 @@
...
+  children,
+  ...props
+}) => {
+  const theme = useTheme();
+
+  return (
+    <div
+      className={classNames?.overlayWrapper}
+      css={{ marginTop: 3 }}
+      {...omit(props, ['onKeyUp', 'onClick'])}
```
michalisKout, line 31: Why are you removing this props from the div element?
You can assign this to a method to improve the readability or add an explanatory comment.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,199 @@
...
+                value={getDateFormatted(selectedDay.from)}
+              />
+              <TextField
+                rightIcon={<Icon name={'calendarEmpty'} color={'secondary'} />}
+                label={`Date (End)`}
+                {...props}
+                lean={true}
+                value={getDateFormatted(selectedDay.to)}
+              />
+            </div>
```
michalisKout, line 182: you can extract this to a new component to make the existing component more testable and clean.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,199 @@
...
+          firstDayOfWeek: 1,
+          numberOfMonths: isRangePicker ? 2 : 1,
+          disabledDays: disableFutureDates
+            ? [
+                {
+                  after: new Date(),
+                },
+              ]
+            : undefined,
+        }}
```
michalisKout, line 157: You can assign this to a variable to reduce length of the rendered JSX code. I think, it will be easier for someone to read the final rendered JSX.

example: 
```
return <SomeComponent  dayPickerProps={{
          onDayClick: isRangePicker ? handleDayRangeClick : handleDayClick,
          selectedDays: selectedDay as RangeModifier,
          modifiers: isRangePicker ? modifiers : undefined,
          firstDayOfWeek: 1,
          numberOfMonths: isRangePicker ? 2 : 1,
          disabledDays: disableFutureDates
            ? [
                {
                  after: new Date(),
                },
              ]
            : undefined,
        }}/>
```
vs
```
return <SomeComponent dayPickerProps={dayPickerProps}/>
```

michalisKout:
@panvourtsis We can add a "close" button to the top right corner. If I select a start or end date i cannot dismiss the date picker clicking outside of the picker area.

src/components/DatePicker/OverlayComponent/OverlayComponent.tsx
```diff
@@ -0,0 +1,55 @@
...
+    >
+      <div className={classNames?.overlay}>
+        <div css={{ display: 'flex', flexDirection: 'row' }}>
+          {isRangePicker && (
+            <div css={{ borderRight: '1px solid #dfdfdf' }}>
+              {extraOptions.map(option => (
+                <div
+                  key={option.value}
+                  css={optionStyle({ selected: selectedOption === option.value })(theme)}
+                  onClick={() => setSelectedOption(option.value)}
```
mkarajohn, line 41: I'd prefer to see event handlers as `() => { setSelectedOption(option.value) }`, but eh.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,199 @@
...
+    [selectedDay, setSelectedDay, setSelectedOption, dayPickerRef.current]
+  );
+
+  const handleDayClick = useCallback(
+    day => {
+      setSelectedDay({ from: day, to: day });
+      setSelectedOption('custom');
+      dayPickerRef.current?.hideDayPicker();
+    },
+    [setSelectedDay, setSelectedOption, dayPickerRef.current]
```
mkarajohn, line 107: >ESLint: React Hook useCallback has an unnecessary dependency: 'dayPickerRef.current'. Either exclude it or remove the dependency array. Mutable values like 'dayPickerRef.current' aren't valid dependencies because mutating them doesn't re-render the component.(react-hooks/exhaustive-deps)

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,199 @@
...
+      );
+
+      if (aboutToCompleteBothDates) {
+        dayPickerRef.current?.hideDayPicker();
+      }
+
+      setSelectedDay(range);
+      setSelectedOption('custom');
+    },
+    [selectedDay, setSelectedDay, setSelectedOption, dayPickerRef.current]
```
mkarajohn, line 98: >ESLint: React Hook useCallback has a missing dependency: 'daysInitialState'. Either include it or remove the dependency array. Mutable values like 'dayPickerRef.current' aren't valid dependencies because mutating them doesn't re-render the component.(react-hooks/exhaustive-deps)

src/components/DatePicker/DatePicker.tsx
mkarajohn: I'd suggest you enable "Run `eslint --fix` on save", through Webstorm's settings.

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,199 @@
...
+      onChange(selectedDay);
+    }
+  }, [onChange, selectedDay]);
+
+  const handleDayRangeClick = useCallback(
+    day => {
+      const aboutToCompleteBothDates = selectedDay.from && !selectedDay.to;
+      const range = DateUtils.addDayToRange(
+        day,
+        // @TODO problem with the library as they define RangeModifiers as not null or undefined when they use it like this
```
mkarajohn, line 86: >they define RangeModifiers as not null or undefined when they use it like this

Is it possible to adhere by their rule?

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,199 @@
...
+import useTheme from 'hooks/useTheme';
+import DayPickerInput from 'react-day-picker/DayPickerInput';
+import { DateUtils, RangeModifier, DayPickerInputProps } from 'react-day-picker';
+import 'react-day-picker/lib/style.css';
+import dayjs from 'dayjs';
+
+import TextField from '../TextField';
+import Icon from '../Icon';
+import { wrapperStyle } from '../TextField/TextField.style';
+import { flex } from '../../theme/functions';
```
mkarajohn, line 14: Import can be shortened to `theme/functions`

mkarajohn (changes requested):

mkarajohn:
As a general comment, I'd like to see full keyboard support. I don't know how easy or if it's even feasible with this lib, but it would be nice. (after putting focus on one of the fields and the overlay appears I'd expect keyboard focus to switch to there)

src/components/DatePicker/DatePicker.tsx
panvourtsis: i don't care if its going to be relative or absolute path to be honest as both serve the same purpose and it will work at the end :)  but we can decide one way or another

src/components/DatePicker/DatePicker.tsx
panvourtsis: no because in their documentation in order to start the date range you must define it as undefined if you don't need a value at start. Their type has always a date, they don't accept ```undefined``` or ```null```

panvourtsis:
> As a general comment, I'd like to see full keyboard support. I don't know how easy or if it's even feasible with this lib, but it would be nice. (after putting focus on one of the fields and the overlay appears I'd expect keyboard focus to switch to there)

For this although i made a progress yesterday needs more time and thinking due to the extra params that we are applying (Last 7 days etc), time that will be introduced and navigation on range. So i suggest to split it in another PR so i will open an issue with it

panvourtsis:
> @panvourtsis We can add a "close" button to the top right corner. If I select a start or end date i cannot dismiss the date picker clicking outside of the picker area.

There are some buttons by design for this which i forgot and implement now so please re-check

michalisKout (approved):

mkarajohn:
I noticed a behaviour that doesn't sit right with me:

After you have filled in a start and end date, if you open the randge picker again, it starts with the month of the end date.
image


I would expect that it would start with the month of the start date 
image

mkarajohn (changes requested):
See previous comment

panvourtsis:
@mkarajohn you are right same for me after you said it so i changed it

mkarajohn (approved):

PR #69, opened 2020-09-10
fix: datepicker with value and disabledDates from outside

panvourtsis:
This PR introduce to the datepicker 2 new properties
* disableDates
* value

There would be always a case that we will need to disable some dates in the calendar in order to avoid user clicking on them. This is very common and also there is a need for it in a current task in Video hunter.

Also value is also need if someone want to control the value with an external state.

You can also find two new stories that shows both cases.

## Screenshots
disabled dates
Screenshot 2020-09-10 at 5 42 51 PM

value defined in the component
Screenshot 2020-09-10 at 5 42 56 PM

### DIFF
--- a/src/components/DatePicker/DatePicker.stories.mdx
+++ b/src/components/DatePicker/DatePicker.stories.mdx
@@ -3,6 +3,7 @@ import { withKnobs, boolean, array, select, text } from '@storybook/addon-knobs'
 import DatePicker from './DatePicker';
 import Stack from '../storyUtils/Stack';
 import PresentComponent from '../storyUtils/PresentComponent';
+import dayjs from 'dayjs';
 
 <Meta title="Design System|DatePicker" component={DatePicker} />
 
@@ -39,5 +40,42 @@ Implementation of the simple DayPicker
 <Preview>
   <Story name="DayPicker">
     <DatePicker />
+  </Story>
+</Preview>
+
+# DayPicker with disabled dates
+
+Some disable dates functionality
+
+<Preview>
+  <Story name="DayPicker with disabled dates">
+    <DatePicker
+      disableDates={[
+        dayjs()
+          .add(1, 'day')
+          .toDate(),
+        { daysOfWeek: [0, 6] },
+      ]}
+    />
+  </Story>
+</Preview>
+
+# DayPicker with predefined values
+
+This datepicker has a predefined initial value that can be used both for initial definition and to handle values externally
+
+<Preview>
+  <Story name="DayPicker with predefined values">
+    <DatePicker
+      value={{
+        to: dayjs()
+          .add(11, 'day')
+          .toDate(),
+        from: dayjs()
+          .add(1, 'day')
+          .toDate(),
+      }}
+      disableDates={[{ daysOfWeek: [0, 2, 3, 4, 5, 6] }]}
+    />
   </Story>
 </Preview>
--- a/src/components/DatePicker/DatePicker.stories.mdx
+++ b/src/components/DatePicker/DatePicker.stories.mdx
@@ -1,9 +1,14 @@
 import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
-import { withKnobs, boolean, array, select, text } from '@storybook/addon-knobs';
+import { withKnobs, select } from '@storybook/addon-knobs';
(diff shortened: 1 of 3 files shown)


src/components/DatePicker/DatePicker.style.ts
```diff
@@ -57,17 +57,22 @@ export const datePickerStyles = ({ isRangePicker }: { isRangePicker?: boolean })
...
   .DayPicker-Day--start:not(.DayPicker-Day--outside) {
     background-color: #dfdfdf;
     color: #000;
   }
 
+  .DayPicker-Day--selected.DayPicker-Day--disabled:not(.DayPicker-Day--outside) {
+    opacity: 0.5;
+  }
+
   .DayPicker-Day:hover {
```
burbuli8ra, line 76: I would suggest to update the css selector here as well from `.DayPicker-Dat:hover` to `.DayPicker-Day:not(.DayPicker-Day--disabled):hover`.
This way we can avoid giving the user a false sense of selecting a date that is disabled.
Also, I see that when selecting a day at the moment, the default outline is shown... I think it should be best if we disabled this as well by adding an extra rule, like:
```
.DayPicker-Day:focus {
  outline: none;
}
```

but if we want to keep the outline, since when an active day is selected the selection table collapses and it doesn't affect the UX so much, I would suggest disabling it at least for disabled dates. 😄

src/components/DatePicker/DatePicker.tsx
```diff
@@ -17,6 +18,10 @@ export type Props = {
   isRangePicker?: boolean;
   /** A callback to return user selection */
   onChange?: (date: DateRange) => DateRange;
+  /** Option to disable some dates */
+  disableDates?: Modifier[];
+  /** Value to define if needed an initial state or to handle it externally */
+  value?: RangeModifier;
```
burbuli8ra, line 24: Since we are adding a `value` prop that makes our component controlled from uncontrolled, shouldn't we also add an `onChange` function as prop that would give to our users the ability to update the value from where our datepicker is being used?
Not sure about my suggestion, because I don't know what's the case for this enhancement, but if we are ok for now with only the addition of `value` prop, I believe that we should consider adding an `onChange` function as an enhancement in the future, it would be really useful!

src/components/DatePicker/DatePicker.stories.mdx
```diff
@@ -41,3 +42,40 @@ Implementation of the simple DayPicker
     <DatePicker />
   </Story>
 </Preview>
+
+# DayPicker with disabled dates
```
burbuli8ra, line 46: Probably you are about to hate me... and I am really, really sorry for that, but since the `disabledDates` property of the `DatePicker` component has many options that can be given as standalone properties or can be combined, I think it should be useful to add all the available options inside the **Docs** tab, along with a combination of options as an example, or even add them as a knob. 😞 (sorry)

src/components/DatePicker/DatePicker.style.ts
panvourtsis: i agree for both comments

src/components/DatePicker/DatePicker.tsx
panvourtsis: onChange is there

src/components/DatePicker/DatePicker.stories.mdx
panvourtsis: as this has many options but defined in the property type as Modifier so that the developer can easily know what to add. If i put it with knobs means that there is an interest for someone outside the FE team to play with the component.

How do you suggest to put it on knobs?

src/components/DatePicker/DatePicker.tsx
burbuli8ra: Missed it! 🙈

src/components/DatePicker/DatePicker.stories.mdx
burbuli8ra: I totally agree for the use outside FE on the Docs part!
But if you consider that the provided options are from the Datepicke library and someone inside the team might not know how to specify them, it would be useful to have them inside the code for reference!
Also, for the knob, I was thinking maybe a dropdown with some examples, so that a designer as well could play with it and have a more clear image on the capability of the library used.

burbuli8ra (approved):

PR #71, opened 2020-09-14
fix: datepicker on change combined with value was not working properly

panvourtsis:
This PR is a hot fix solving an issue that when `onChange` and `value` prop where defined and you were trying to handle the component from parent there was a infinite loop on the component rendering causing trouble.

Now as you can see we detect external `value` changes and handle them internally in order to handle both `stateless` and `statefull` cases.

A new function also created to merge and simplify functionality of range and day pick

### DIFF
--- a/src/components/DatePicker/DatePicker.tsx
+++ b/src/components/DatePicker/DatePicker.tsx
@@ -1,12 +1,12 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import React, { useState, useCallback, useEffect, useRef } from 'react';
+import React, { useState, useCallback, useRef, useEffect } from 'react';
 import { datePickerStyles } from './DatePicker.style';
 import useTheme from 'hooks/useTheme';
 import DayPickerInput from 'react-day-picker/DayPickerInput';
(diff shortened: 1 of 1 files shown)


src/components/DatePicker/DatePicker.tsx
```diff
@@ -80,10 +80,24 @@ const DatePicker: React.FC<Props> = ({
...
-      onChange(selectedDay);
+    if (value.from && value.to) {
+      setSelectedDay(value);
     }
-  }, [onChange, selectedDay]);
+  }, [value]);
+
+  const handleCalendarValueChange = useCallback(
+    (range, hideDatePicker = false) => {
+      setSelectedDay(range);
```
michalisKout, line 90: Here you can create a stronger cohesion between the updates by adding a second argument (callback fn) which will run after the update of the selectedDay. 

example:
`setSelectedDay(range, () => setSelectedOption('custom')); `

PS: Although, I didn't find a way to reproduce it on typescript as the SetStateAction accepts 1 argument.
```type SetStateAction<S> = S | ((prevState: S) => S);```

michalisKout (approved):

PR #79, opened 2020-09-16
[TextField] Fix width to fit the parent container

burbuli8ra:
This PR fixes the width of TextField that is now not fully expanded into its parent container.

Behaviour before fix:
Screenshot 2020-09-16 at 2 53 47 PM

Behaviour after fix:
Screenshot 2020-09-16 at 2 52 41 PM


panvourtsis:
can you fix the input to get all the space needed because this happens now

image

panvourtsis (approved):

PR #81, opened 2020-09-17
introduce custom label to datepicker

panvourtsis:
This PR extends the functionality of the Datepicker and adds a new prop to handle a custom label for the input/inputs.

This way we can add anything we want as we do on the `TextField` component. The prop is optional and defaults to the label of the design `Date`

### Screenshot
image

### More
I also extended the knobs functionality so we can play with these values


Christos-Zacharopoulos (approved):
Maybe we need different labels in each input. (From - To ).  :nerd_face:

Issue #78, opened 2020-09-16
TextField New Styles


burbuli8ra:
This issue refers to the 1st bullet of issue #55 

We need to enhance the TextField component with the following styles:
- [x] Filled
- [x] Outlined
- [x] Elevated

We also need to modify the styling of all mentioned styles above for **focus** state

Designs on Abstract: [here](https://app.abstract.com/projects/96260c1d-f169-4306-ab88-2472d8e457fd/branches/e8803d0d-0c59-4eeb-9d4e-1cae0df12a08/commits/latest/files/d7b356b2-3f84-47de-8d32-8b811b5c4590/layers/9FFE5B59-098C-47A1-B457-14FE07A845B8?collectionId=ed3a1a60-78e2-4d01-a8bd-fd4e8d6356cd&collectionLayerId=672046ce-c06c-4677-bd2a-3dc5f27bdcc6)

PR #82, opened 2020-09-17
[TextField] Add new styles and focus state

burbuli8ra:
This PR introduces the following:

**1.** New styles of the TextField component:
- Filled
- Outlined
- Elevated

**2.** Some minor fixes on the existing default style (filled).
The fixes applied are the following:
- Replacement of border with box-shadow inset (for avoiding resizing of component on state change)
- Fix styling of label when the input value is empty (should have font-weight normal)
- Fix styling of error state (should have a light red background)
- Fix disabled state (should have a white layer with opacity)

**3.** Addition of focus state for all components

_Note 1: All existing stories have been updated to show all the available styles along with the lean option._
_Note 2: Minor fixes to the colors have been made to match the designs._

Closes issue #78 

Designs on Abstract: [here](https://app.abstract.com/projects/96260c1d-f169-4306-ab88-2472d8e457fd/branches/e8803d0d-0c59-4eeb-9d4e-1cae0df12a08/commits/latest/files/d7b356b2-3f84-47de-8d32-8b811b5c4590/layers/9FFE5B59-098C-47A1-B457-14FE07A845B8?collectionId=ed3a1a60-78e2-4d01-a8bd-fd4e8d6356cd&collectionLayerId=672046ce-c06c-4677-bd2a-3dc5f27bdcc6)

Since the changes are applied on several states of the component, I believe that the best way to see the changes is through storybook, so I will not be uploading any screenshots here.

### DIFF
--- a/src/components/TextField/TextField.style.ts
+++ b/src/components/TextField/TextField.style.ts
@@ -1,5 +1,5 @@
 import { css, SerializedStyles } from '@emotion/core';
-import { rem, transparentize } from 'polished';
+import { rem } from 'polished';
 import { Props } from './TextField';
 import { Theme } from '../../theme';
 
@@ -38,17 +38,11 @@ const wrapperStyleSwitch = (theme: Theme, lean?: boolean, error?: boolean, style
     case 'filled':
     default:
       return `
-        background-color: ${
-          error ? transparentize(0.85, theme.palette.error[400]) : theme.palette.flat.lightGray[200]
-        };
-        box-shadow: inset 0 0 0 1px ${
-          error ? theme.palette.error[400] : theme.palette.flat.lightGray[200]
-        };
+        background-color: ${error ? '#fdf2f2' : '#f5f5f5'};
+        box-shadow: inset 0 0 0 1px ${error ? theme.palette.error[400] : '#f5f5f5'};
         
         &:focus-within {
-          box-shadow: inset 0 0 0 1px ${
-            error ? theme.palette.error[400] : theme.palette.flat.lightGray[200]
-          },
+          box-shadow: inset 0 0 0 1px ${error ? theme.palette.error[400] : '#f5f5f5'},
             0px 2px 6px 0px rgba(67, 67, 67, 0.15);
         }
       `;
--- a/src/components/TextField/TextField.style.ts
+++ b/src/components/TextField/TextField.style.ts
@@ -96,6 +96,7 @@ export const textFieldStyle = ({ label, leftIcon }: Props) => (
   width: fill-available;
 
   label {
+    color: ${theme.palette.flat.lightGray[500]};
     left: ${leftIcon ? '1.9rem' : 'inherit'};
   }
 `;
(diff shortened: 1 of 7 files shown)


src/components/TextField/TextField.style.ts
```diff
@@ -1,32 +1,85 @@
 import { css, SerializedStyles } from '@emotion/core';
-import { rem, transparentize } from 'polished';
-import { Theme } from '../../theme';
 import { Props } from './TextField';
+import { rem } from 'polished';
 
+const wrapperStyleSwitch = (theme: Theme, lean?: boolean, error?: boolean, styleType?: string) => {
+  if (lean) {
```
burbuli8ra, line 6: Maybe adding `lean` to the styling options, should be a good idea.
Since I am not aware of which projects are using the lean version of the component, my suggestion is to add it to the component's `styleType` available options and add a todo to remove the `lean` property when we have replaced it with `styleType="lean"` to our projects!

src/components/TextField/TextField.style.ts
```diff
@@ -1,32 +1,82 @@
...
+      box-shadow: inset 0 0 0 1px transparent;
+    `;
+  }
+
+  switch (styleType) {
+    case 'outlined':
+      return `
+        background-color: white;
+        box-shadow: inset 0 0 0 1px
+          ${error ? theme.palette.error[400] : '#dfdfdf'};
```
burbuli8ra, line 19: Colors that do not match our color palette, have been replaced according to designs!

src/components/TextField/TextField.style.ts
panvourtsis: lean is to bypass anything that styles the input as a general but you want the mechanism - check datepicker

src/components/storyUtils/TextFieldShowcase/TextFieldShowcase.tsx
```diff
@@ -5,23 +5,32 @@ import TextField from 'components/TextField';
 import { Props as TextFieldProps } from 'components/TextField/TextField';
 
 type Props = {
+  disabled?: boolean;
   errorMsg?: React.ReactNode | string;
   label?: string;
+  lean?: boolean;
   status?: string;
+  styleType?: 'filled' | 'outlined' | 'elevated';
```
panvourtsis, line 13: i suggest move those to a global type as it will be used from the select, datepicker etc etc

panvourtsis (approved):
The component looking really good i also tried datepicker as well but i suggest create a new ticket for datepicker and implement this prop there as well which will be pretty easy.

src/components/TextField/TextField.style.ts
Christos-Zacharopoulos: Here the palette colors should be used. 

If the designs have different colors than the ones in palette we should discuss it with the design team to either change the palette yet again or update the designs.

burbuli8ra: Sure @Christos-Zacharopoulos !
We will sync with the design team in order to find a solution that fits both sides.
For now I am leaving the components as they are and will add a todo comment (as you suggested) in order to keep track of the future required changes!

src/components/storyUtils/TextFieldShowcase/TextFieldShowcase.tsx
burbuli8ra: @panvourtsis I will add it as a type in `palette` file if you 're ok with it!

PR #85, opened 2020-09-18
Expose components

Christos-Zacharopoulos:
This PR fixes the following:

Not all the created components were exposed from index.


panvourtsis (commented):
I am not sure that `Label` must be exposed

panvourtsis (approved):

PR #88, opened 2020-09-21
Feat. Change the way special grays are created. 

Christos-Zacharopoulos:
This PR introduces a special way to handle the LightGray and DarkGray colors.

All the colors are created by using shade or tint to a base color with 25% percentage. 

The grays start from the darkest one, which is #010101 and get lighter with 8% percentage on each step, except the last one that's pure white.

specialGrays


panvourtsis (changes requested):
i expect to see the description of this in a documentation

panvourtsis (approved):

PR #90, opened 2020-09-23
fix: remove useTheme that was used only for css functions

panvourtsis:
## Description

This PR does a cleanup on the components that were using the `useTheme` hook only to pass the theme property into the css functions. This is not required as for object styles this is being handled by emotion. Furthermore i also removed the `SerializedStyles` type from styles that does nothing for the css functions

## Screenshot

Only code changes

## Test Plan

Storyshots tests are updated based on the css objects returned

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/src/components/Breadcrumb/Breadcrumb.tsx
@@ -5,7 +5,6 @@ import isEmpty from 'lodash/isEmpty';
 import uniqueId from 'lodash/uniqueId';
 import * as React from 'react';
 import { Link } from 'react-router-dom';
-import useTheme from '../../hooks/useTheme';
 import { breadcrumbLinkStyles, breadcrumbStyles } from './Breadcrumb.style';
 import BreadcrumbCollapsed from './BreadcrumbCollapsed/BreadcrumbCollapsed';
 import BreadcrumbItem from './BreadcrumbItem/BreadcrumbItem';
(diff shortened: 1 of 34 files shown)


burbuli8ra (commented):
By removing `SerializedStyles` ESLint throws a warning that the return type of the function `(theme: Theme) => css\`\`` is missing.
Either than that PR looks great! 👏

Christos-Zacharopoulos (changes requested):
There are still some usages of useTheme hook (DatePickInput, Menu, RenderRowOrNestedRow etc)

panvourtsis:
> There are still some usages of useTheme hook (DatePickInput, Menu, RenderRowOrNestedRow etc)

i removed it when its not used. For example on DatePickInput the css prop that get an array is not returning a theme therefore i use the hook. Only object css prop returns theme property

panvourtsis:
> By removing `SerializedStyles` ESLint throws a warning that the return type of the function `(theme: Theme) => css``` is missing.
> Either than that PR looks great! 👏

hmmmm @burbuli8ra why i dont see it ? where is that ?

panvourtsis:
@burbuli8ra you were right about the type so i put them back. This is because we expect theme and that the function will be handled by emotion but the function itself doesn't know that is going to be used only by emotion

Christos-Zacharopoulos (approved):

burbuli8ra (approved):

PR #99, opened 2020-10-13
Fix/table and checkbox

panvourtsis:
## Description

This PR fix two bugs that i found on the table with the checkbox. 

Checkbox issue was that although we defined centering items through flex the parent was not a `flex` item so it wasn't working. This was due to the fact that when the parent was `flex` then the hover was not working correctly. Now i moved the hover to the pseudo :before.

Table issue was that when the checkbox was shown on some tables the width was smaller and because of the padding and the different width of the header cell was moving.

[designs](https://share.goabstract.com/e4bb166a-21e9-421f-b27d-c0f5215baeaf?collectionLayerId=ed92aefd-8e0d-4e49-b457-459dcb21ee65&mode=design)

## Screenshot

before
Screenshot 2020-10-13 at 12 38 18 PM

after
Screenshot 2020-10-13 at 12 21 40 PM
Screenshot 2020-10-13 at 12 21 27 PM

## Test Plan

Tests are only made through snapshots

### DIFF
--- a/src/components/CheckBox/CheckBox.style.ts
+++ b/src/components/CheckBox/CheckBox.style.ts
@@ -7,6 +7,7 @@ export const wrapperStyle = ({ disabled }: Props) => (): SerializedStyles => css
   opacity: ${disabled ? 0.3 : 1};
   justify-content: center;
   align-items: center;
+  display: flex;
 `;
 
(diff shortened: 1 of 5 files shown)


panvourtsis:
@Orfium/front-end reminder for the PR

burbuli8ra (approved):

PR #101, opened 2020-10-21
feat(datepicker): adds date format option for date pick input

### LINKED ISSUE #100: allow datepicker formating on input
Right now there is a specific format on the input of the datepicker. We must allow the user to define this format e.g `dd/mm/yyyy`

michalisKout:
## Description

This PR adds the ability to the user to change the date format of the date picker based on the list of localized formats in [dayjs](https://day.js.org/docs/en/display/format).
## Screenshot
**Example of a new date format:**
Screenshot 2020-10-21 at 1 45 40 PM

**List of localized formats:**
Screenshot 2020-10-21 at 1 46 04 PM

resolves #100

### DIFF
--- a/src/components/DatePicker/DatePickInput/DatePickInput.tsx
+++ b/src/components/DatePicker/DatePickInput/DatePickInput.tsx
@@ -11,6 +11,8 @@ import { formFieldStyles } from '../../../theme/palette';
 import localizedFormat from 'dayjs/plugin/localizedFormat';
 import localeData from 'dayjs/plugin/localeData';
 import { getLocaleFormat } from '../../../utils/helpers';
+import 'dayjs/locale/en-gb';
+import 'dayjs/locale/en';
 
 type Props = {
   isRangePicker: boolean;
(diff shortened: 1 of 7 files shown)


panvourtsis:
something that i would like to see as well is the localized format since the products are being used in both EU and US. This will add value as options of `dd/mm/yyyy` or `mm/dd/yyyy` will seem odd to people from the opposite side. What do you say can we add it here?

michalisKout:
> something that i would like to see as well is the localized format since the products are being used in both EU and US. This will add value as options of `dd/mm/yyyy` or `mm/dd/yyyy` will seem odd to people from the opposite side. What do you say can we add it here?

I agree that the format of the date may seem odd to people from the EU / US.

Wouldn't be more clear if the application that will eventually use the date picker has the responsibility to provide the correct date format to the date picker component based on the location of the client - (i18n)?
This way, we can keep our component's date format abstract and flexible.

Let me know what you think.

panvourtsis:
if we had such thing like a prop `localized: boolean` and the format will still be as you made it i don't see if it will have a problem. If we move such logic to the project on the other hand we might have the same logic splitted in different projects. So i think it is value to have it inside and with splitted logic will also be abstract as well.

michalisKout:
> if we had such thing like a prop `localized: boolean` and the format will still be as you made it i don't see if it will have a problem. If we move such logic to the project on the other hand we might have the same logic splitted in different projects. So i think it is value to have it inside and with splitted logic will also be abstract as well.

Ok great, we can split the logic by adding a new boolean prop, as you mentioned above. 😄  👍

michalisKout:
**Status Update**: After discussing with @panvourtsis, we decided to add a new functionality to the Date Picker component regarding the system's locale format.

To achieve this, I used the [locale extension](https://day.js.org/docs/en/plugin/locale-data) method from the [dayjs](url) library.

Note: _An alternative way to achieve this would be to use the `.toLocaleString()` method. Although, It seems pretty unstable having to manipulate a string value that may change in the future._

src/components/DatePicker/DatePickInput/DatePickInput.tsx
```diff
@@ -6,28 +6,46 @@ import { flex } from '../../../theme/functions';
...
 type Props = {
   isRangePicker: boolean;
   selectedDay: DateRange;
   inputLabel: string;
   /** Style of input field */
   styleType: formFieldStyles;
+  dateFormatOverride?: DateFormatType;
 } & DayPickerInputProps;
 
+(function InitDatePickerLocaleFormat(usLocale: string, euLocale: string) {
```
panvourtsis, line 26: this will define the whole app format in a bigger environment so i suggest to move it to the themeProvider level. What do you say?

src/components/DatePicker/DatePickInput/DatePickInput.tsx
```diff
@@ -6,28 +6,46 @@ import { flex } from '../../../theme/functions';
...
-  const getDateFormatted = React.useCallback(
-    (date: Date | undefined) => (date ? dayjs(date).format('MM/DD/YYYY') : ''),
-    []
+  const formatDate = (date: Date | undefined) => {
+    return date ? dayjs(date).format(getLocaleFormat(dateFormatOverride)) : '';
+  };
+
+  const getDateFormatted = React.useCallback(formatDate, []);
+
+  console.log(
```
panvourtsis, line 47: reminder

src/components/DatePicker/DatePickInput/DatePickInput.tsx
michalisKout: I agree. We can even create a new ts file (`initLocale.ts`) and import it to the index.tsx of the app.

panvourtsis: totally agree 💯

panvourtsis (approved):

panvourtsis:
lgtm 🎊  good job !

PR #110, opened 2020-11-19
feat: new release config for style and docs

panvourtsis:
## Description

Style changes with or without scope should trigger a patch release. This PR changes the config that this could happen.
Also i added a docs functionality for the documentation that we will apply for also a patch version so we can keep track.

## Screenshot

No screenshot necessary

## Test Plan
No tests are necessary

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,7 +1,13 @@
 {
     "branches": "master",
     "plugins": [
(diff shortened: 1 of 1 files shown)


Christos-Zacharopoulos (approved):
We should also update the contribution docs.

PR #113, opened 2020-11-19
Feature/update palette

panvourtsis:
## Description

This PR introduce a new functionality to introduce a new utils in the theme. The first thing in these utils will be a function to simplify to get colors from our palette.

### Implementation
This function will be a standalone one and the same will be in the theme and could be accessed like `theme.utils.getColor`

This function has 3 different scopes 
`normal` -> that will point in the lower level color/types like primary, success, info, warning etc
`text` -> that will point to the text palette
`flat` -> which is the default value that doesn't need to define it. This points to the flat colors AKA our system palette

### Usage
```
theme.utils.getColor('darkGray', 400) // for flat
theme.utils.getColor('primary', 100, 'text') // for text palette
theme.utils.getColor('warning', 400, 'normal') // for warning in the type/colors
```

## Screenshot

No need.

## Test Plan

Tests for new `getColor` function that returns the correct color.


panvourtsis:
@Orfium/front-end reminder for this

Christos-Zacharopoulos:
Some stories seem to have broken with the introduction of the getColor function and the changes of this PR. (Icon stories for example).

Christos-Zacharopoulos (approved):
:tada:

michalisKout (approved):

Christos-Zacharopoulos (approved):

Issue #58, opened 2020-09-03
Notifications


panvourtsis:
Notifications will handle two actions if found in props
Will support close functionality
Will *not* handle when it will show - this will be handled by the user, component will remain dummy

Designs can be found [here](https://app.abstract.com/projects/96260c1d-f169-4306-ab88-2472d8e457fd/branches/e8803d0d-0c59-4eeb-9d4e-1cae0df12a08/commits/latest/files/d7cb764d-aa66-4fd6-90dd-4804917a3ed3/layers/336D3D69-30DA-4DD5-B855-5F18BBCE962A?collectionId=69d4231d-574e-4e8c-89e8-70a68ba6edbb&collectionLayerId=1532083c-ec58-4e0c-a1af-c54419f0b0bc) for all notifications

- [x] inline - would be a component that would not be absolute positioned so that the user can show it as inline everywehre
- [x] banner - will be an absolute positioned component with top|bottom - right|left position
- [ ] toast - will be a absolute positioned component with top|bottom - right|left position - Also toggled functionality
- [ ] modals - will be a floating component with centered position

PR #121, opened 2020-11-30
Feature/notifications banner

kostasdano:
## Description

This is the second part of the implementation of the issue #58

- Banner Notification is implemented, with all possible types, variations and positioning

## Screenshot

- Banner Notifications [Abstract link](https://app.abstract.com/projects/96260c1d-f169-4306-ab88-2472d8e457fd/branches/master/commits/latest/files/d7cb764d-aa66-4fd6-90dd-4804917a3ed3/layers/336D3D69-30DA-4DD5-B855-5F18BBCE962A?collectionId=ae1fabbe-9c71-457e-8959-f7fee7411487&collectionLayerId=45bd2b6b-0673-4f71-a107-89161bf8b103)
banners

- Positioning

banners-gif

## Test Plan

Test scenarios:

- Banner Container with 2 Inline Notifications renders correctly

- Banner Container Inline Notifications primaryCTAs work properly when clicked

- Banner Container Inline Notifications closeCTAs work properly when clicked


src/components/Notification/Notification.stories.mdx
```diff
@@ -73,6 +74,57 @@ Inline Notification
...
 
+# Banner Notification
+
+Banner Notification
+
+<Preview>
+  <Story name="Banner Notifiction" parameters={{ decorators: [withKnobs] }}>
+    <PresentComponent name="" width={1024}>
+      <BannersContainer
+        top={boolean('top', false)}
```
michalisKout, line 85: Kostas kudos for your first PR in Ictinus! I would suggest creating a more 'user-friendly' approach as described in issue #58 (with top|bottom - right|left position). 
As a user of this container, I would expect to pass a string in the property that will ensure the position of the banner. 

For example:
Container's types:
```
interface Props  {
 position: "top-right" | "top-left" | "bottom-left" | "bottom-right";
}
```

Invocation:
`<BannersContainer position="top-right" />`

The position depends on a combination of two rules described in the above example. This way you can ensure that the user will never miss a position placement.

kostasdano: Actually that's a great idea. A bug of the current implementation is that the user may insert weird props combinations (for example `<BannersContainer top bottom/>` so with your combinations system idea we avoid such cases

src/components/Notification/subcomponents/NotificationsContainer/NotificationsContainer.tsx
```diff
@@ -0,0 +1,28 @@
...
+type Positions = 'top-right' | 'top-left' | 'bottom-left' | 'bottom-right';
+
+type Props = {
+  /** Notifications Container children */
+  children: React.ReactNode[];
+  /** Notifications Container position */
+  position: Positions;
+};
+
+const positionToBooleans = (position: Positions) =>
```
michalisKout, line 15: To avoid the complexity of the nested conditions you can export the position logic to an object.

Example: 
```
const positionOptions = {
 'top-right': [true, false, false, true],
'top-left': [true, false, false, true],
'bottom-left': [false, true, true, false],
'bottom-right': [false, true, false, true]
};

//in the Container you will have...
const currentPosition = positionOptions[position];
return <div css={notificationsContainer(currentPosition)}>{children}</div>;
```

or even better using a factory function
```
const getPositionStyle = (styleLiteral: string) => {
  return css`
    ${styleLiteral}
  `;
};
const positionOptions = {
  'top-right': `top:0; right: 0;`,
  'top-left': `top:0; left: 0;`,
  'bottom-left': `bottom:0; left: 0;`,
  'bottom-right': `bottom:0; right: 0;`,
};

use it like this...
export const notificationsContainer = (currentPositon: Positions) => (
  theme: Theme
): SerializedStyles => css`
   position: absolute;
 .... 
  ${getPositionStyle(positionOptions[currentPositon])}
`;
```

Christos-Zacharopoulos (approved):
:tada:

src/components/Notification/Notification.test.tsx
```diff
@@ -57,3 +58,94 @@ describe('Inline Notification', () => {
...
+
+    const primaryCTALabel1 = await findByText(data1.primaryCTALabel);
+    expect(primaryCTALabel1).toBeTruthy();
+
+    const primaryCTALabel2 = await findByText(data1.primaryCTALabel);
+    expect(primaryCTALabel2).toBeTruthy();
+  });
+
+  test('Notifications Container Banner Notifications primaryCTAs work properly when clicked', async () => {
+    const primaryCTA1 = jest.fn();
```
michalisKout, line 108: It's great to see tests here! Kudos Kostas!
Is there any specific reason for the double checks? 
I believe that you can just use one mocked function and one <Notification /> component to test all its functionalities.

kostasdano: Well, given the fact that Notification Container can contain multiple Notification components I thought it would be nice to test that multiple primaryCTAs and closeCTAs work properly and aren't affecting each other (eg: each CTA click does not interfere with other CTAs etc). Well I'm not really sure if that's possible and if that's a necessary check; I'd like to hear your opinion on that approach.

michalisKout: Your logic is correct, having many children components may affect things. Although, I think this has to do with UI things and not the functionality of the component.
Since the click event will be emitted inside the Notification component you don't have to worry if it will affect its siblings (this is an effect of event bubbling in JS).

michalisKout (approved):

Issue #114, opened 2020-11-23
Line Chart


panvourtsis:
## Description

Create a new component for showing a line chart in our design system. This component will provide both single line and multiple line chart. 

## Design
The VH project [design](https://share.goabstract.com/f22ffb16-5644-4d49-bcdf-247e9c248bdc) that will implement these components 

<img width="334" alt="Screenshot 2020-11-23 at 12 42 29 PM" src="https://user-images.githubusercontent.com/6433679/99953224-a9f35c80-2d89-11eb-8f63-83e5515149e8.png">
<img width="329" alt="Screenshot 2020-11-23 at 12 41 47 PM" src="https://user-images.githubusercontent.com/6433679/99953230-abbd2000-2d89-11eb-840c-4c7b0d55842a.png">
<img width="334" alt="Screenshot 2020-11-23 at 12 41 53 PM" src="https://user-images.githubusercontent.com/6433679/99953234-ac55b680-2d89-11eb-8ec7-efe9337eb2b9.png">

## Implementation

Need to use the **[recharts](https://recharts.org/en-US)** library to implement the line chart as show above in the screenshots. [recharts github](https://github.com/recharts/recharts)

- Line chart must take data for a line that will demonstrate X and Y axis
- Line chart component must take multiple lines (data).
- Each line must have a unique color either defined or randomly picked from our palette
- Line chart will have the option to show the titles of the lines at the bottom left. e.g () Hidden () Claimed
- On Point hover/click a popup must show with a description of that point if provided
- On Multiple Line Point hover/click a popup must show with a description of that point if provided and the point of other lines 

## Tests

- X, Y axis are shown correctly
- Count lines based on data passed
- Hover/Click on points works

PR #122, opened 2020-12-01
Feature/Line-chart

SVonKarajann:
## Description

A LineChart component for showing both single line and multiple line charts.
[ictinus issue](https://github.com/Orfium/orfium-ictinus/issues/114)
[design](https://app.abstract.com/projects/27ff87f3-129d-472d-a51e-6d9ce0fa3ef0/branches/8b3fb2c5-d747-4f39-8a32-d2552b0f89c3/collections/d7cd386c-9a02-489f-a55b-4d5ff2e1f023)

## Screenshot

chart2


src/components/LineChart/components/CustomTooltip/CustomTooltip.style.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export const tooltipLiStyle = () => (theme: Theme) => {
+  return {
+    listStyleType: 'none',
+    color: theme.palette.white,
+    width: '100%',
+    display: 'flex',
+    justifyContent: 'space-between',
+    height: '16px',
```
kostasdano, line 38: instead of using px, try using the rem() function from polished and the standard spacing from ictinus (where possible)

eg: 16px -> rem(theme.spacing.md). This can apply for all pixel spacings from 4px to 32px with rounding when needed. For bigger spacings you can use only the rem function (eg. rem(200))

panvourtsis:
please @SVonKarajann revisit the comment here and fix accordingly

src/components/LineChart/components/CustomTooltip/CustomTooltip.style.ts
SVonKarajann: thanks for the comment. done!

SVonKarajann:
fixed any comment-related issue

SVonKarajann:
@kostasdano @michalisKout just a reminder, please review asap

src/components/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,125 @@
...
+    return (
+      <ResponsiveContainer aspect={1.5}>
+        <AreaChart data={data} margin={{ top: 10, right: 20, left: 20, bottom: 50 }}>
+          {children}
+        </AreaChart>
+      </ResponsiveContainer>
+    );
+  }
+
+  // for testing purposes only !!!
```
michalisKout, line 112: We should make the chart's size responsive for real cases too.

src/components/LineChart/story.utils.ts
```diff
@@ -0,0 +1,57 @@
+export const initData = [
```
michalisKout, line 1: These are mocked data, not utils. Can we please move them to stories.mdx file?

src/components/LineChart/LineChart.tsx
SVonKarajann: Not sure if i understand your comment correctly. We return `<ResponsiveContainer> 
<AreaChart/> 
</ResponsiveContainer>` only for real cases. On test cases we just return `<AreaChart/>` because rechart's ResponsiveContainer component was not working well with our tests.

src/components/LineChart/story.utils.ts
SVonKarajann: I use initData in LineChart.test.ts file as well, guess i should move them there as well?

src/components/LineChart/LineChart.tsx
michalisKout: Testing a library is necessary as soon as the library does not cover a case by its tests. 

Besides that, you created a dependency on a system variable (NODE_ENV). We may face side effects by using this component on a third-party project.

Maybe we should consider removing the dummy component.

src/components/LineChart/story.utils.ts
michalisKout: or you can move them in a file "mockedData.ts" 😃

michalisKout (approved):

michalisKout:
@panvourtsis is this going for v1 or v2?

panvourtsis:
> @panvourtsis is this ready for v1 or v2?

it doesnt use something for v2

PR #125, opened 2020-12-02
Feature Top Application Bar

### LINKED ISSUE #115: Top Bar Navigation
## Description
Depends on #126 #127 to finish
Create a new component for top bar that will hold basic functionality for all the projects

## Design
Cases of the navbar are demonstrated in this [design](https://share.goabstract.com/d050c9c1-16b2-4370-abcc-37e07b87df75?mode=design&sha=e0b3d557d7564040d702696635dd90445d0bad78)

image

## Implementation

The navbar component will have several implementations. The weird and difficult part here is that the components that its using are almost the component in the system but not quite. So we will not use the internal components here. We will build something new that will be used only by navbar. These components will be

- custom search input
- custom drop down with selections
- custom drop down with selection (user)

For more reference check Video Hunter project.

Need to implement
- Background color for navbar defined as prop
- Burger Button on the left that will trigger a callback passed by the parent (to open a menu maybe)
- Create a logo placeholder
- Search Input will be the component based on #126 
- - This will have a onChange callback to notify the parent of any change
- The client dropdown will be a custom component area that the dev could use to show anything 
- Dropdown (Menu) - the component will be based on #127 
- - callback to notify onChange
- - dropdown with label, Icon and options (1 or many options) - User Case

## Tests

- Search input callback is being called once - onClick
- Burger Button is being called once - onClick
- dropdown shown
- - all values shown correctly (labelm icon, options) based on the data passed
- dropdown selection triggered parent callback

michalisKout:
## Description
This PR implements the Top Application Bar with its features as described in issue #115.
resolves #115 

## Screenshot

Screenshot 2021-01-05 at 6 03 14 PM

## Test Plan

Snapshot testing


src/components/TopAppBar/TopAppBar.style.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const topAppBarWrapper = ({ bgColor }: StyleProps) => (theme: Theme): SerializedStyles => css`
+  ${flexCenter};
+  position: relative;
+  justify-content: space-between;
+  background-color: ${theme.utils.getColor(bgColor.type, bgColor.variant)};
+  color: ${pickTextColorFromSwatches(bgColor.type, bgColor.variant)};
+  height: ${DEFAULT_NAVBAR_HEIGHT}px;
+  padding: ${theme.spacing.sm} ${theme.spacing.md};
```
panvourtsis, line 20: this padding must change to `padding: 0 ${theme.spacing.md};` as the components inside has the `sm`. If you put this the height of the topbar exceeds the 62px which is the design

src/components/TopAppBar/TopAppBar.tsx
```diff
@@ -0,0 +1,63 @@
...
+      role="banner"
+      aria-label="Top Application Banner"
+      css={Styles.topAppBarWrapper({ bgColor })}
+    >
+      <div css={Styles.mainSection}>
+        <SidebarMenuIcon onMenuIconClick={onMenuIconClick} />
+        <LogoWrapper logoIcon={logoIcon} />
+        <div css={Styles.searchWrapper}>
+          <TextField
+            label={searchLabel}
```
panvourtsis, line 38: label here must change to placeholder e.g `placeholder={searchLabel}`. The label does the animation material like which we dont want at the top bar

panvourtsis (commented):
few comments. Other than that seem great! Awesome work! 🎊

panvourtsis (approved):

PR #130, opened 2020-12-03
Feature Card

### LINKED ISSUE #116: Card
## Description

There is a need for a new Card in the system that will be a placeholder for anything to wrap in. Basically this will be a styled wrapper for children.

## Design
The cards
image

## Implementation
The card should have one prop - elevated that should take `01,02,03,04` as options.

The options will be based on the below which you can find also [here](https://share.goabstract.com/f07c6215-b593-4b1e-803d-d5dea0650402?collectionLayerId=df2df991-c012-400a-9146-3545d942a0d8&mode=design
) 
Elevation must be calculated base on a function that will be globally accessable from others as this will be applied to other components if needed
image

## Testing
Unit test on function to return the correct elvation
Snapshot Testing for the component

kouloumos:
## Description

This PR implements the Card with its features as described in issue #116 
Resolves #116 

## Screenshot
The card is a wrapper for children and thus can take different shapes

Screenshot 2020-12-04 at 7 30 53 PM

## Test Plan

- Snapshot testing for the component

### DIFF
--- a/src/components/Card/Card.style.ts
+++ b/src/components/Card/Card.style.ts
@@ -1,0 +1,12 @@
+import { css, SerializedStyles } from '@emotion/core';
+import { Props } from './Card';
+import { Theme } from '../../theme';
+import { Elevation } from 'index';
+
+export const cardStyle = ({ elevated }: Props) => (theme: Theme): SerializedStyles => css`
+  ${elevated && cardElevation(theme, elevated)};
+`;
+
+export const cardElevation = (theme: Theme, elevated: keyof Elevation): SerializedStyles => css`
+  box-shadow: ${theme.elevation[elevated]};
+`;
--- a/src/components/Card/Card.stories.mdx
+++ b/src/components/Card/Card.stories.mdx
@@ -1,0 +1,36 @@
+import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
+import { select } from '@storybook/addon-knobs';
+import CardShowcase from '../storyUtils/CardShowcase';
+import Card from './Card';
+
+<Meta title="Design System/Card" component={Card} />
+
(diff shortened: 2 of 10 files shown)


panvourtsis:
according to the design elevatad is optional as it could be a simple card.

src/components/Card/Card.style.ts
```diff
@@ -0,0 +1,13 @@
+import { css, SerializedStyles } from '@emotion/core';
+import { Props } from './Card';
+import { Theme } from '../../theme';
+import { Elevation } from 'index';
+
+export const cardStyle = ({ elevated }: Props) => (theme: Theme): SerializedStyles => css`
+  display: inline-block;
```
panvourtsis, line 7: i am not so sure that changing native block to inline-block is a good idea. I would rather like to see the parent making the adjustments, something like the below e.g based on your card showcase

```
const CardShowcase = () =>
  elevations.map(elevation => {
    return (
      <div key={elevation} css={{ marginBottom: '40px', '& > *': { height: 100, width: 749 } }}>
        <Card elevated={elevation}>Card with {elevation} elevation</Card>
      </div>
    );
  });
```

kouloumos: @panvourtsis So the logic is that the parent element will wrap around the card to give the proper dimensions? Because we defined the card as a wrapper for other elements, that's why I thought that the card's height&width will be set according to the elements in it, but I couldn't do it without `display: inline-block` because otherwise width was always at 100% of the page.

With that inner `<div/>` i tried to replicate a component with set dimensions, that the card would wrap around and hence the card would get the size of the component (aka inner div). Is that the wrong way of thinking about this?

panvourtsis (approved):

PR #136, opened 2020-12-08
refactor/menu

### LINKED ISSUE #127: Menu: update size and allow custom icons
## Description 
This will change the menu component and add additional functionalities for custom icons (left/right) and introduce a new sizing prop to define the size of the component

## Design
Component will be based on the select design and can be found [here](https://share.goabstract.com/4336243a-14aa-41b6-81a7-dacd51e2ad82?collectionLayerId=888fdb90-094b-46db-b94c-e0ad2df1f284&mode=design) 
image

## Implementation
We need to update the current implementation based on the select designs but there will be no label option

- add prop to pass a new component at the left
- update prop to pass the component at the right (maybe will be a different icon than dots)
- update the old padding logic with icon based on the new design flow
- Add avatar component for left icon
- change color based on what is passed

panvourtsis:
## Description

This PR introduce a refactored menu for ictinus based on the ticket #127. This menu will be part of the top bar #125

In order to do this i created the Avatar component as well as this is required on the menu.

closes #127 
depends on #134 

## Screenshot

Screenshot 2020-12-16 at 8 06 09 PM
Screenshot 2020-12-16 at 8 29 42 PM

## Test Plan
Storyshots


src/components/Avatar/Avatar.style.ts
```diff
@@ -0,0 +1,63 @@
+import { css, SerializedStyles } from '@emotion/core';
+import { Theme } from '../../theme';
+import { rem } from 'polished';
+import { flex } from '../../theme/functions';
+import { colorShades, flatColors, pickTextColorFromSwatches } from '../../theme/palette';
+import { AvatarShapes, AvatarSizes } from './Avatar';
+
+const sizeBasedOnProp = (size: AvatarSizes) => {
```
michalisKout, line 8: Why don't we use an object here? 

example:
```
const sizeBasedOnProp = {
  'sm': rem(40),
  'md': rem(46),
  'xs': rem(24),
}

sizeBasedOnProp[size] || defaultSize

```

Maybe I've missed something in the business logic here.

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,70 @@
...
+      className={className}
+      css={avatarStyle({
+        shape,
+        size,
+        fill: calculatedColor.color,
+        fillShade: calculatedColor.shade,
+      })}
+    >
+      {src && <img src={src} />}
+      {!src && !children && iconName && (
```
michalisKout, line 58: Do we have the same case here as the TextField? Should we let the user decide how to pass his icon/image?

Also, give an alt to the image 😄

src/components/Menu/Menu.style.ts
```diff
@@ -83,30 +31,30 @@ export type MenuOptions = {
...
+  box-shadow: ${theme.elevation['02']};
+  display: flex;
+  flex-direction: column;
+  justify-content: space-evenly;
+  text-align: center;
+  border-radius: ${rem(4)};
+  z-index: 1;
+  & > button {
+    padding: ${rem(8)} 0;
+    height: ${rem(48)};
```
NikosCha, line 52: the buttons are stretched when lots of options are added.
image

adding `min-height:  ${rem(48)};` seems to solve the problem

src/components/Avatar/Avatar.style.ts
panvourtsis: same thing isn't it? also my solution offers the `default` return that in TS will be avoided by in not in JS :)

src/components/Avatar/Avatar.tsx
panvourtsis: this is an `iconName` for sure. It will never be an external icon here unless its a src of an avatar

src/components/Menu/Menu.style.ts
NikosCha: Another problem that arises by adding lots of options is that lots of them exceeds the parent container and are not visible 
image 
image

NikosCha (commented):
- According to the designs, the scrollbar is different.
image 
- I think we dont need a scrollbar when there are not items to scroll. 
image

src/components/Avatar/Avatar.style.ts
michalisKout: Actually, it isn't. 

It's similar but having many `if` conditions here means that you will have more performance cost (asking every time for the correct case) than just picking the correct object's reference. 

Also, it's a more expandable way by adding a new prop with a value than adding an if condition.

Although, if your solution serves its purpose, I'm totally fine 😃 .

panvourtsis: i will change it to a switch case as it will serve better for our case

panvourtsis:
> * According to the designs, the scrollbar is different.
>   image
> * I think we dont need a scrollbar when there are not items to scroll.
>   image

good catch this was a but with `overflow-y: scroll;` that now is fixed

src/components/Menu/Menu.style.ts
panvourtsis: i think i fixed what you are mentioning here but please recheck the latest version

panvourtsis:
image
as this will be a blocking issue for your please recheck @michalisKout

src/components/Menu/Menu.style.ts
NikosCha: Works fine! thank you

NikosCha (commented):
My comments are fine now, just one thing. The scrollbar is still different to designs. If this is not the case I will approve it.

panvourtsis:
> My comments are fine now, just one thing. The scrollbar is still different to designs. If this is not the case I will approve it.

the scrollbar is something that we need to fix globally not on component basis as this will add inconsistencies through the projects. So the design team either must have a strict proposal for this (and we handle it on a side task) or we stick with OS one.

NikosCha (approved):

michalisKout:
Great work @panvourtsis! Should we also include the arrow rotation on the menu toggling? I think we missed that.

We could also open a new issue for the scrollbar waiting for the design team to provide us a solution.

Christos-Zacharopoulos:
> We could also open a new issue for the scrollbar waiting for the design team to provide us a solution.

This is a great package to have the same styling on different browsers ( I am looking at you Firefox )
https://www.npmjs.com/package/react-custom-scrollbars

panvourtsis:
> > We could also open a new issue for the scrollbar waiting for the design team to provide us a solution.
> 
> This is a great package to have the same styling on different browsers ( I am looking at you Firefox )
> https://www.npmjs.com/package/react-custom-scrollbars

thanks for this @Christos-Zacharopoulos and @michalisKout this needs work and i am creating a new ticket for it

michalisKout (approved):

PR #144, opened 2020-12-21
Fix/Notification Minor Fixes

kostasdano:
## Description

Minor Fixes for **InlineNotification** and **NotificationsContainer**:

- Icon and Filling props optionality fixed
- NotificationsContainer is now in fixed position, not absolute


mkarajohn (commented):
When a boolean prop is optional it's convention that if it has a default value it's `false`, so that when you set, for example `withIcon` it actually sets it to true.

panvourtsis (approved):
LGTM 🎊

PR #154, opened 2021-01-08
Feature/modal

Christos-Zacharopoulos:
## Description

In this PR we introduce the modal component. 

Resolves issue #58 part 4.

Designs [here](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System?node-id=1163%3A7455)
## Screenshot

Modal

## Test Plan

Snapshot testing and also tests that check:

- Modal renders correctly
- Modal closes
- Modal actions are called correctly
- Modal displays all content correctly


michalisKout:
Nicely done Christos. 

Although I think the ModalContent should be a prop of the Modal. At the moment, we can use the ModalContent only if we render it as child of Modal.

It is an implementation detail that a user of DS does not have to know about.

I would suggest to change it to something like this:
<Modal fixedContent />

```
if(fixedContent){
 ...display <ModalContent />
}

...display children
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,50 @@
...
+  /** The data test id if needed */
+  dataTestId?: TestId;
+};
+
+const Modal: React.FC<Props> = ({ open = false, onClose, dataTestId, children }) => {
+  if (!open) return null;
+
+  return (
+    <div
+      css={backgroundContainer()}
```
michalisKout, line 25: You can change all the functions with no variables to const as variables are much "cheaper" than function calls.

src/components/Modal/ModalContent/ModalContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+  heading,
+  message,
+  primaryCTA,
+  primaryCTALabel,
+  secondaryCTA,
+  secondaryCTALabel,
+  dataTestId,
+}) => {
+  return (
+    <div
```
michalisKout, line 45: You can change this to `dialog` to make it semantically correct.

src/components/Modal/ModalContent/ModalContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+            type={'primary'}
+            filled={false}
+            onClick={secondaryCTA}
+            size={'md'}
+            dataTestId={'modal-content-secondaryCTA'}
+          >
+            {secondaryCTALabel}
+          </Button>
+        )}
+        {Boolean(primaryCTA && primaryCTALabel) && (
```
michalisKout, line 64: Since you are reusing this condition you can assign it to a variable.

example:
```
const shouldDisplayCTA = Boolean(primaryCTA && primaryCTALabel);
```

src/components/Modal/ModalContent/ModalContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+      css={modalContentContainer()}
+      data-testid={generateTestDataId('modal-content', dataTestId)}
+    >
+      {Boolean(label) && <p css={labelContainer()}>{label}</p>}
+      <h5 css={headingContainer()}>{heading}</h5>
+      <p css={messageContainer()}>{message}</p>
+      <div css={actionsContainer()}>
+        {Boolean(secondaryCTA && secondaryCTALabel) && (
+          <Button
+            type={'primary'}
```
michalisKout, line 55: Should this type be 'secondary'?

src/components/Modal/ModalContent/ModalContent.tsx
mkarajohn: Don't. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#browser_compatibility

michalisKout: `role="dialog"` then 😉

src/components/Modal/ModalContent/ModalContent.tsx
Christos-Zacharopoulos: This condition is not re-used.

src/components/Modal/ModalContent/ModalContent.tsx
Christos-Zacharopoulos: If you check the designs, both buttons have type `primary` by design but the only the left has filled set to `true`.

After discussion with design team, this is because the left button will be mainly used for actions like canceling an action  and the right one is the "important" that confirms an action.

Christos-Zacharopoulos:
> Nicely done Christos.
> 
> Although I think the ModalContent should be a prop of the Modal. At the moment, we can use the ModalContent only if we render it as child of Modal.
> 
> It is an implementation detail that a user of DS does not have to know about.
> 
> I would suggest to change it to something like this:
> ```
> if(fixedContent){
>  ...display <ModalContent />
> }
> 
> ...display children
> ```

I am not sure I like this approach. Then the user should pass the pros of ModalContent to Modal as well which will be undefined if the content is something else.

 And as it is, the Modal wrapper can (and will) be reused elsewhere also (ex. [file uploading component](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System?node-id=483%3A6106) ).

michalisKout:
> > Nicely done Christos.
> > Although I think the ModalContent should be a prop of the Modal. At the moment, we can use the ModalContent only if we render it as child of Modal.
> > It is an implementation detail that a user of DS does not have to know about.
> > I would suggest to change it to something like this:
> > ```
> > if(fixedContent){
> >  ...display <ModalContent />
> > }
> > 
> > ...display children
> > ```
> 
> I am not sure I like this approach. Then the user should pass the pros of ModalContent to Modal as well which will be undefined if the content is something else.
> 
> And as it is, the Modal wrapper can (and will) be reused elsewhere also (ex. [file uploading component](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System?node-id=483%3A6106) ).

The user will pass these props anyway. 
Having the modalContent as a prop in Modal will indicate that there is a fixed content option.
Otherwise, the modal users may miss the fact that there is a ready to use component for this type of usage.

 example: 
 ```
<Modal fixedContent={ ctaPrimaryLabel: '', ....} />
 ```

I think we may need some changes if we gonna use it for file uploading since the top modal bar is slightly different.

We can also take it offline if you want.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,66 @@
...
+import ClickAwayListener from '../utils/ClickAwayListener';
+import ModalContent, { Props as ModalContentProps } from './ModalContent/ModalContent';
+
+export type Props = {
+  /**  If true, the modal is open. Defaults to false. */
+  open: boolean;
+  /** Callback fired when the component requests to be closed. */
+  onClose: () => void;
+  /** If true, modal will use default ModalContent and contentProps should be ModalContentProps. Defaults to false */
+  fixedContent?: boolean;
```
michalisKout, line 18: I think you can remove this and only use the contentProps but it's optional, so I will leave it to you.

michalisKout (approved):

PR #165, opened 2021-01-20
Feature/textfield v2

panvourtsis:
## Description
In this PR we introduce the new TextField styling based on the v2 components.

As this is the final version components that are using this are going to change such as
- Topbar (search input)
- Dropdown (another pr will include this - #168)

Also at this point the dev can have access to the actual ref of the input so it can be used to any external lib like formik etc.

Designs can be found [here](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System?node-id=1012%3A49306)

## Screenshot

image
image

## Test Plan
- Design QA
- Storyshots


src/components/TextField/TextField.tsx
```diff
@@ -81,16 +81,16 @@ const TextField = React.forwardRef<HTMLInputElement, Props>(
       </React.Fragment>
     );
 
-    const IconWrapper: FC = ({ children }) => (
-      <div css={iconWrapperStyle({ label, rightIcon, leftIcon })}>{children}</div>
+    const IconWrapper: FC<{ iconPosition: 'left' | 'right' }> = ({ children, iconPosition }) => (
```
michalisKout, line 84: optional: You can expose a type for iconPosition to reuse it in styles.

michalisKout:
lgtm 👍🏻

src/components/TextField/TextField.tsx
panvourtsis: it was only one and didnt want to mess for one property type and redirect the user back and forth to check the reference.

panvourtsis:
> ~lgtm 👍🏻 ~
> 
> Screenshot 2021-01-21 at 4 01 01 PM
> 
> @panvourtsis
> 
> Actually, I just spotted this issue in the Select component.

this is implemeted in another PR as stated in this PR description #168

NikosCha (approved):

PR #168, opened 2021-01-21
Feature/dropdown v2

panvourtsis:
## Description
This PR implements the designs for the simple dropdown. 

This dropdown will be able to get options and show them. Each option must have a specific shape. The component can take a default value from the options in order to start with that. Also each select is filterable.

React-Select library also has been removed as it makes things complicated.

Multiple dropdown will be splitted to another PR as the checkbox that will use is under some minor fixes

depends on #165 

## Screenshot
image
image

## Test Plan
- storyshots
- design qa


src/components/DatePicker/DatePickInput/DatePickInput.tsx
```diff
@@ -1,5 +1,6 @@
 import dayjs from 'dayjs';
 import * as React from 'react';
+import omit from 'lodash/omit';
 import { DayPickerInputProps } from 'react-day-picker';
 import useTheme from '../../../hooks/useTheme';
```
gsmetak, line 5: Is a blank line required above this one, to separate the imports?

gsmetak (approved):

src/components/DatePicker/DatePickInput/DatePickInput.tsx
panvourtsis: not quite sure what do you mean here? can you send an example?

kouloumos: maybe he is referring to a styling that separates imports from external libraries, internal etc?

panvourtsis: actually i have such thing in my notes to implement to all projects. But as doing this manually is actually us shooting at our feet i will import a solution to do this with a rule

PR #170, opened 2021-01-25
Bugfix/select fix

panvourtsis:
## Description
This PR solves several issues from select listed below

- defaultValue when change the value selected didnt change
- arrow click when click to focus on input on select and open menu
- menu position change from relative to absolute so it can hover over elements
- TextField small size was 44px rather than 46px

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -1,13 +1,14 @@
 /** @jsx jsx */
 import * as React from 'react';
+import { useEffect, useMemo, useRef } from 'react';
 import useTheme from '../../hooks/useTheme';
 import TextField from '../TextField';
 import Icon from '../Icon';
 import { Props as TextFieldProps } from '../TextField/TextField';
 import { jsx } from '@emotion/core';
 import ClickAwayListener from '../utils/ClickAwayListener';
-import { useEffect, useMemo } from 'react';
 import { menuStyle, optionStyle } from './Select.style';
+import useCombinedRefs from '../../hooks/useCombinedRefs';
 
 export type SelectOption = {
   value: string | number;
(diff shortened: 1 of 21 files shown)


src/components/Select/Select.tsx
```diff
@@ -65,9 +72,12 @@ const Select = React.forwardRef<HTMLInputElement, Props>(
           size={20}
           name={open ? 'chevronLargeUp' : 'chevronLargeDown'}
           color={theme.utils.getColor('lightGray', 500)}
+          onClick={() => {
+            inputRef?.current?.focus();
```
michalisKout, line 76: I think it is more UX-friendly to toggle only the list of options instead of focusing when you click on the arrow icon. 
This way you can also avoid the complexity of multiple refs.

example:
```
setOpen(!open);
```

michalisKout (approved):
LGTM 👌

michalisKout:
@panvourtsis  can we include a max-height CSS property? At the moment if we render many options the dropdown height will be huge. 

Can we introduce two new props maxHeight & maxWidth?

Finally, please check the `select-with-icon-and-size` case in the storybook, seems broken. 

Sorry for the delayed comment after my approval ✌️.

michalisKout (approved):

michalisKout:
Dropdown items don't seem to have ellipsis in longer-than-expected texts.


Screenshot 2021-01-26 at 11 27 19 AM

I believe that additional style properties needed in `optionStyle`.

.
example:
```
    max-width: 400px;
    white-space: nowrap;
    overflow: hidden;
    text-overflow: ellipsis;
```

Screenshot 2021-01-26 at 11 40 49 AM

panvourtsis:
> Dropdown items don't seem to have ellipsis in longer-than-expected texts.
> 
> Screenshot 2021-01-26 at 11 27 19 AM
> 
> I believe that additional style properties needed in `optionStyle`.
> 
> .
> example:
> 
> ```
>     max-width: 400px;
>     white-space: nowrap;
>     overflow: hidden;
>     text-overflow: ellipsis;
> ```
> 
> Screenshot 2021-01-26 at 11 40 49 AM

this is now fixed

michalisKout (approved):

PR #173, opened 2021-01-27
[DS-68] Feature/text area

panvourtsis:
## Description
This PR introduce a new text area component. In order to accomplish same styling rules accross all of our components based on the text field i split the style logic to be centralized under a wrapper. This wrapper now will be used in text field and text area.

Designs can be found [here](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System?node-id=67%3A31)

## Screenshot
image

## Test Plan
QA design
Snapshots

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -1,7 +1,6 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import * as React from 'react';
-import { FC } from 'react';
+import React, { FC, InputHTMLAttributes } from 'react';
 import { iconWrapperStyle, inputStyle } from './TextField.style';
 import Label from '../Label';
 import Icon from '../Icon';
@@ -26,7 +25,9 @@ export type Props = {
   onInput?: React.EventHandler<any>;
 } & TextInputWrapperProps;
 
-const TextField = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
+type InputProps = Omit<InputHTMLAttributes<HTMLInputElement>, 'size'>;
+
+const TextField = React.forwardRef<HTMLInputElement, Props & InputProps>((props, ref) => {
   const {
     id = undefined,
     rightIcon = null,
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -2,50 +2,18 @@
 import { jsx } from '@emotion/core';
 import * as React from 'react';
 import { FC } from 'react';
-import {
-  errorMsgStyle,
-  iconWrapperStyle,
-  inputStyle,
-  textFieldStyle,
-  wrapperStyle,
-} from './TextField.style';
+import { iconWrapperStyle, inputStyle } from './TextField.style';
 import Label from '../Label';
 import Icon from '../Icon';
-import { formFieldStyles } from 'theme/palette';
-import { DEFAULT_SIZE } from '../../utils/size-utils';
-import useTheme from '../../hooks/useTheme';
+import { DEFAULT_SIZE } from 'utils/size-utils';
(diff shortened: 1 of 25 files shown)


panvourtsis:
image
@Orfium/front-end @kouloumos @NikosCha

src/components/TextField/TextField.tsx
```diff
@@ -56,94 +24,65 @@ export type Props = {
   onKeyDown?: React.KeyboardEventHandler<HTMLTextAreaElement | HTMLInputElement>;
   /** Callback fired when the `input` value typed is changed */
   onInput?: React.EventHandler<any>;
-};
+} & TextInputWrapperProps;
```
Christos-Zacharopoulos, line 27: As discussed this is not in the scope of the TextArea component but should be resolved.

Consider changing the type of the `leftIcon` and `rightIcon` from  `string | JSX.Element | null;` to `JSX.Element | null`

Christos-Zacharopoulos (approved):

src/components/utils/TextInputWrapper/TextInputWrapper.style.ts
```diff
@@ -0,0 +1,173 @@
...
+        content: '';
+        background-color: rgba(255, 255, 255, 0.15);
+        height: 100%;
+        width: 100%;
+        position: absolute;
+        left: 0;
+        top: 0;
+        z-index: 1;
+      }
+      * {
```
michalisKout, line 90: These rules may create UI inconsistencies in the future. We should refactor CSS to avoid using `*` (select all operators) and `!important` tags.

src/components/TextField/TextField.tsx
```diff
@@ -56,94 +24,65 @@ export type Props = {
...
+              label={label}
+              required={required}
+              animateToTop={Boolean(rest.value)}
+            />
+          )}
         </div>
-        {hintMsg && status !== 'normal' && (
-          <div css={errorMsgStyle({ status })}>{hintMessageToShow}</div>
+        {rightIcon && !locked && <IconWrapper iconPosition={'right'}>{rightIcon}</IconWrapper>}
+        {locked && (
```
michalisKout, line 74: You can move this condition inside IconWrapper's body.

example: 

```
 const IconWrapper: FC<{ iconPosition: 'left' | 'right' }> = ({ children, iconPosition }) => (
    locked && <div css={iconWrapperStyle({ iconPosition })}>{children}</div>
  );
```

src/components/utils/TextInputWrapper/TextInputWrapper.style.ts
panvourtsis: because this case is in disabled or locked state this should replace all. Its way easier and readable in my opinion this way rather than having 100 if elses here and there. What do you think?

src/components/TextField/TextField.tsx
panvourtsis: actually this makes it harder for me to understand 😄 , i dont fully agree with this `shouldDisplay` functionality

src/components/utils/TextInputWrapper/TextInputWrapper.style.ts
michalisKout: After a close look at this, I think the logic operator here should change to `(disabled || locked) && ...` the parenthesis will change the behavior of the styles.

Since we manage the colors and not the users, we can remove `!important` tags and create rules for specific elements.

example:
```
     > input,textarea {
        color: ${theme.utils.getColor('lightGray', 600)};
        fill: ${theme.utils.getColor('lightGray', 600)};
      }
```

NikosCha (commented):
Please fix the conflicts

michalisKout (approved):

NikosCha (approved):

PR #176, opened 2021-01-28
fix(textfield): fixes input props causing a typescript error

gsmetak:
(Possibly) Fixes TS errors that would emerge when trying to add generic HTML input attributes to the component (e.g. maxLength). 
Also changes types in other components that use the TextField, in order to accommodate the new changes.

### DIFF
--- a/src/components/DatePicker/DatePickInput/DatePickInput.tsx
+++ b/src/components/DatePicker/DatePickInput/DatePickInput.tsx
@@ -10,6 +10,7 @@ import { wrapperStyle } from '../../TextField/TextField.style';
 import { DateFormatType, DateRange } from '../DatePicker';
 import { formFieldStyles } from '../../../theme/palette';
 import { getLocaleFormat } from '../../../utils/helpers';
+import { InputHTMLAttributes } from 'react';
 
 // TODO: Need to fix this (TextField onChange prop)
 const ON_CHANGE_MOCK = () => {};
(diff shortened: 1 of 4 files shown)


src/components/DatePicker/DatePickInput/DatePickInput.tsx
```diff
@@ -10,6 +10,7 @@ import { wrapperStyle } from '../../TextField/TextField.style';
 import { DateFormatType, DateRange } from '../DatePicker';
 import { formFieldStyles } from '../../../theme/palette';
 import { getLocaleFormat } from '../../../utils/helpers';
+import { InputHTMLAttributes } from 'react';
```
panvourtsis, line 13: you can change import from react in line 2

src/components/TextField/TextField.tsx
```diff
@@ -1,7 +1,7 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
 import * as React from 'react';
-import { FC } from 'react';
+import { FC, InputHTMLAttributes } from 'react';
```
panvourtsis, line 4: same here please fix import

src/components/Select/Select.tsx
```diff
@@ -87,13 +87,13 @@ const Select = React.forwardRef<HTMLInputElement, Props>(
         <TextField
           onFocus={() => setOpen(true)}
           rightIcon={rightIconRender}
-          onKeyDown={e => {
+          onKeyDown={(e: any) => {
```
panvourtsis, line 90: please use ts KeyboardEvent instead of any

src/components/Select/Select.tsx
```diff
@@ -87,13 +87,13 @@ const Select = React.forwardRef<HTMLInputElement, Props>(
...
             // if backspace
             if (e.keyCode === 8) {
               setInputValue(emptyValue);
             }
           }}
-          onInput={e => {
+          onInput={(e: any) => {
```
panvourtsis, line 96: same here dont use any please

panvourtsis (approved):
LGTM 🎊

PR #190, opened 2021-02-22
Feature/breakpoints - responsive util

panvourtsis:
## Description

This PR introduce a new mechanism to determine the window size and return based on that what DS size is currently allowed.

[figma design ](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System?node-id=119%3A3)

The sizes right now allowed are the following.
`[
  '1920',
  '1440',
  '1366',
  '1200',
  '1024',
  '970',
  '750',
  '480',
  '320',
]`
We keep in mind that our responsive system will have a `mobile first` approach which means that the lowest to highest size will be served. A small example of what this means in our system is this screenshot that shows that we are currently supporting `tablet 970` and down.

image

## Screenshot
figma DS needs

image

## Test Plan
unit tests on the hook implementation

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "lodash": "^4.17.19",
     "pluralize": "^8.0.0",
     "polished": "^3.4.4",
+    "react-media": "^2.0.0-rc.1",
     "recharts": "^1.8.5"
   },
   "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@svgr/webpack": "^5.4.0",
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.0.1",
+    "@testing-library/react-hooks": "^5.0.3",
     "@types/events": "^3.0.0",
     "@types/jest": "^25.1.0",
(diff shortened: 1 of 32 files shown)


package.json
```diff
@@ -81,6 +82,7 @@
     "lodash": "^4.17.19",
     "pluralize": "^8.0.0",
     "polished": "^3.4.4",
+    "react-media": "^2.0.0-rc.1",
```
Christos-Zacharopoulos, line 85: :scream_cat:

src/components/storyUtils/Breakpoints/Breakpoints.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import useBreakpoints, { queriesKeys } from 'hooks/useBreakpoints';
+
+const MyComponent = () => {
```
Christos-Zacharopoulos, line 4: Now I am just being annoying, I know, but could you change this to Breakpoints showcase for consistency?

Christos-Zacharopoulos (approved):
Also fix the empty file but in general it looks good to me!

:tada:

PR #193, opened 2021-02-23
feat(select): extend select functionality

whosyourtaco:
Implement async search functionality for large amounts of options.

Add Loader component and loading status.

Add minimum characters to search for option.

Add a text-highlight option for searched term appearances in available options.

Add respective stories for showcasing.

Add isSearchable prop.

### DIFF
--- a/src/components/Loader/Loader.style.tsx
+++ b/src/components/Loader/Loader.style.tsx
@@ -1,0 +1,58 @@
+import styled from 'styled-components';
+
+export const Container = styled.div`
+  align-self: center;
+`;
+
+export const Dots = styled.div`
+  position: relative;
+  width: 6px;
+  height: 6px;
+  border-radius: 5px;
+  background-color: #cfcfcf;
+  color: #cfcfcf;
+  animation: dotFlashing 0.7s infinite linear alternate;
+  animation-delay: 0.5s;
+
+  &::after,
+  &::before {
+    content: '';
+    display: inline-block;
+    position: absolute;
+    top: 0;
+  }
+
+  &::before {
+    left: -10px;
+    width: 6px;
+    height: 6px;
+    border-radius: 5px;
+    background-color: #cfcfcf;
+    color: #cfcfcf;
+    animation: dotFlashing 0.7s infinite alternate;
+    animation-delay: 0s;
+  }
+
+  &::after {
+    left: 10px;
+    width: 6px;
+    height: 6px;
+    border-radius: 5px;
+    background-color: #cfcfcf;
+    color: #cfcfcf;
+    animation: dotFlashing 0.7s infinite alternate;
+    animation-delay: 0.7s;
+  }
+
+  @keyframes dotFlashing {
+    0% {
+      background-color: #e7e7e7;
+    }
+    50% {
+      background-color: #dbdbdb;
+    }
+    100% {
+      background-color: #cfcfcf;
+    }
+  }
+`;
--- a/src/components/Loader/Loader.style.tsx
+++ b/src/components/Loader/Loader.style.tsx
@@ -1,18 +1,31 @@
 import styled from 'styled-components';
 
+export const getDotsLayout = (delay: number, animation: string, left?: number) => `
+    left: ${`${left}px` || ''};
+    width: 6px;
+    height: 6px;
+    border-radius: 5px;
+    background-color: #cfcfcf;
+    color: #cfcfcf;
+    animation: ${animation};
+    animation-delay: ${delay}s;
+`;
+
 export const Container = styled.div`
   align-self: center;
 `;
 
 export const Dots = styled.div`
   position: relative;
-  width: 6px;
-  height: 6px;
-  border-radius: 5px;
-  background-color: #cfcfcf;
-  color: #cfcfcf;
-  animation: dotFlashing 0.7s infinite linear alternate;
-  animation-delay: 0.5s;
+  ${getDotsLayout(0.5, 'dotFlashing 0.7s infinite linear alternate')}
+
+  &::before {
+    ${getDotsLayout(0, 'dotFlashing 0.7s infinite alternate', -10)}
+  }
+
+  &::after {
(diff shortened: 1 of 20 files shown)


whosyourtaco:
Some info about the implementation:

- Decided against using onKeyDown event, because it fires for _every_ key pressed (shift, ctrl, alt etc.).
- Decided against using onKeyPress event, even though it fires only for characters, because the event it exposes is KeyboardEvent which doesn't contain a reference to the component's value.
- Only choice after was the onChange and that's why I used that.
- Decided on using a prop for the `isAsync` flag, because the `asyncSearch` prop is optional, meaning I would have to check if it's undefined either way if a default value wasn't given. This makes the component more semantically correct IMO, since you can immediately understand if it's the async version or not.
- Any suggestions about the debounce() part of the implementation are very very (**_very_**) welcome

panvourtsis:
@whosyourtaco you can change this to a normal PR

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Container, Dots } from './Loader.style';
+
+type Props = {
+  type?: 'linear' | 'spinner' | 'dots';
```
panvourtsis, line 5: allow only dots for now

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
panvourtsis, line 1: can we add a storybook story for this ?

src/components/Loader/Loader.tsx
whosyourtaco: Alright!

src/components/Loader/Loader.tsx
whosyourtaco: Sure! On it

src/components/Loader/Loader.style.tsx
```diff
@@ -0,0 +1,58 @@
+import styled from 'styled-components';
+
+export const Container = styled.div`
+  align-self: center;
+`;
+
+export const Dots = styled.div`
+  position: relative;
+  width: 6px;
```
michalisKout, line 9: Since the next lines are identical you can export them to a function and invoke the exported function here.

example: 
```
const getDotsLayout = (delay, animation, left) => `
    left: ${left || 'auto'};
    width: 6px;
    height: 6px;
    border-radius: 5px;
    background-color: #cfcfcf;
    color: #cfcfcf;
    animation: ${animation};
    animation-delay: ${delay}s;
    `

styled.div`
${getDotsLayout(0.5, 'dotFlashing 0.7s infinite linear alternate')}

&::before {
     ${getDotsLayout(0, 'dotFlashing 0.7s infinite alternate', -10)}
}  
`
```

P.S. consider also in using the theme's util getColor for the color setup.

src/components/Select/Select.tsx
```diff
@@ -44,45 +55,100 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps>(
...
 
+    const handleOnKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+      // if backspace
+      if (e.keyCode === 8) {
+        setInputValue(emptyValue);
+      }
+    };
+
+    const debouncedOnChange = React.useCallback(
+      debounce(term => {
```
michalisKout, line 98: Maybe we should only debounce side effects on leading action to avoid unexpected behavior or delays, e.g. { leading: true }, what do you think?

src/components/Loader/Loader.style.tsx
whosyourtaco: Sure!

src/components/Select/Select.tsx
whosyourtaco: Nice idea! I'm adding it now

whosyourtaco: These are the results with leading set to true and false. While in theory it sounded better, I believe that the default behavior results in a smoother experience for the user, even though they will have to wait for the debounce timer to end. I can actually reduce the timer from 500ms to 400ms.

leading: true
leading

leading: false
trailing

michalisKout: Thanks for the descriptive comment @whosyourtaco. Let's leave the implementation as is and take it offline for further investigation.

michalisKout (approved):

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type Props = {
+  type?: 'dots';
+};
+
+const Loader: React.FC<Props> = ({ type = 'dots' }) => {
+  const renderLoader = () => {
+    switch (type) {
+      case 'dots':
+        return <Dots data-testid="dots-loading" />;
```
panvourtsis, line 12: there is a util that generate `testid` check utils folder

src/components/Select/Select.tsx
```diff
@@ -45,45 +58,105 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps>(
...
-      setSearchValue('');
+      if (isSearchable) {
+        setSearchValue('');
+      }
       handleSelectedOption(option);
     };
 
+    const handleOnKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+      // if backspace
+      if (e.keyCode === 8) {
```
michalisKout, line 98: This is optional. Comments will be stale as soon as you write them.  Therefore I think it's better to have variables instead of comments for these cases. 

e.x.
```
const isBackspaceKey =  e.keyCode === 8 || e.which === 32;

if(isBackspaceKey) {...}
```

michalisKout (approved):
LGTM 👍🏻 🎉 . In general, I would try to use callbacks instead of logic operators (isAsync, isSearchable). Having too much logic in our component may complicate things in the future.

src/components/Select/Select.tsx
whosyourtaco: This wasn't in the scope of this PR but yeah I totally agree with you. Changing it too

panvourtsis (approved):

PR #194, opened 2021-02-25
Fix: Select props should now have the selected value

elektracodes:
## Description

From 2.7 to 2.8 version the prop that passed the selected value to the parent didn't work. This should fix the issue. Perhaps we should think of adding some unit tests too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "@testing-library/jest-dom": "^5.11.4",
     "@testing-library/react": "^11.0.1",
     "@testing-library/react-hooks": "^5.0.3",
+    "@testing-library/user-event": "^12.7.3",
     "@types/events": "^3.0.0",
     "@types/jest": "^25.1.0",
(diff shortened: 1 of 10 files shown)


whosyourtaco (approved):
LGTM 🎈

package.json
```diff
@@ -28,9 +28,11 @@
     "@storybook/preset-typescript": "^3.0.0",
     "@storybook/react": "^6.1.10",
     "@svgr/webpack": "^5.4.0",
+    "@testing-library/dom": "^7.29.6",
```
panvourtsis, line 31: why we have those to libs in? is it for the userEvent?

src/components/Select/Select.test.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import Select from './Select';
+// import userEvent from '@testing-library/user-event';
```
panvourtsis, line 3: remove unused code

PR #197, opened 2021-03-02
fix(packages): after upgrade fix automatic runtime for pragma

panvourtsis:
## Description

This PR fix the automatic runtime error for pragma 
`Syntax error: pragma and pragmaFrag cannot be set when runtime is automatic.`

I only saw that on latest react version of 16.14.0 and 17. I am not sure why out of a sudden our project start complain about this. Nevertheless i tried removing all the jsx pragma from all the files but there is the problem with the css in the tests that are rendering just a function and not the css prop. The simplest solution is to pass to the files that had the jsx pragma the runtime pragma.

I will revisit for the fix later on an upgrade of the libs for a once and for all solution.

EDITED: i found that this is coming from the latest babel release on 7v

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -1,3 +1,4 @@
+/** @jsxRuntime classic */
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
 import React from 'react';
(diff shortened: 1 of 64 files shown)


Christos-Zacharopoulos (approved):
🎉 

Let's just add an issue in order to revisit this. Did you manage to find an open issue in babel v7 so that we can reference that?

PR #199, opened 2021-03-03
Fix/sidenav

Christos-Zacharopoulos:
## Description

In this PR we make fixes to the side nav component regarding its functionality on desktop (desk1366)

[ticket](https://orfium.atlassian.net/browse/DS-83)

## Screenshot

Capture

## Test Plan

Updated snapshots

### DIFF
--- a/src/components/Drawer/Drawer.style.ts
+++ b/src/components/Drawer/Drawer.style.ts
@@ -3,9 +3,11 @@ import { rem } from 'polished';
 import { transition } from 'theme/functions';
 import { Theme } from '../../theme';
 
-export const drawerContainerStyle = (expanded: boolean, isDesktop: boolean) => (
-  theme: Theme
-): SerializedStyles => css`
+export const drawerContainerStyle = (
+  expanded: boolean,
+  isDesktop: boolean,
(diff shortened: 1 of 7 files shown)


src/components/Drawer/Drawer.style.ts
```diff
@@ -3,9 +3,11 @@ import { rem } from 'polished';
 import { transition } from 'theme/functions';
 import { Theme } from '../../theme';
 
-export const drawerContainerStyle = (expanded: boolean, isDesktop: boolean) => (
-  theme: Theme
-): SerializedStyles => css`
+export const drawerContainerStyle = (
+  expanded: boolean,
+  isDesktop: boolean,
+  isSmallDesktop: boolean
```
elektracodes, line 9: Shouldn't we add for all the small devices instead of just isSmallDesktop?

elektracodes (approved):

src/components/Drawer/Drawer.style.ts
Christos-Zacharopoulos: According to the designs, it's only for small desktop screens that we need to make the Side Navigation expandable on hover.

PR #201, opened 2021-03-03
refactor(loader): change styling from styled-components to emotion

whosyourtaco:
Completely change loader styling to emotion. 

Incorporate theme getColor util function to get the colors from our palette.

### DIFF
--- a/src/components/Loader/Loader.style.tsx
+++ b/src/components/Loader/Loader.style.tsx
@@ -1,30 +1,32 @@
-import styled from 'styled-components';
+import { css, SerializedStyles } from '@emotion/core';
 
-export const getDotsLayout = (delay: number, animation: string, left?: number) => `
-    left: ${`${left}px` || ''};
-    width: 6px;
-    height: 6px;
-    border-radius: 5px;
-    background-color: #cfcfcf;
-    color: #cfcfcf;
-    animation: ${animation};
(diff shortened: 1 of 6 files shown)


src/components/Loader/Loader.style.tsx
```diff
@@ -1,30 +1,32 @@
...
-    background-color: #cfcfcf;
-    color: #cfcfcf;
-    animation: ${animation};
-    animation-delay: ${delay}s;
+import { css, SerializedStyles } from '@emotion/core';
+
+import { Theme } from 'theme';
+
+export const getDotsLayout = (delay: number, animation: string, theme: Theme, left?: number) => css`
+  left: ${`${left}px` || ''};
```
panvourtsis, line 6: this is wrong. This will return `undefinedpx` in all cases. Also the braces are wrong

solution
`left && `${left}px`;`
otherwise it will be undefined and will not set left at all

panvourtsis (changes requested):

src/components/Loader/Loader.style.tsx
whosyourtaco: This is really really wrong, I forgot to change it at some point. It actually worked though, in all cases :P

panvourtsis (approved):

PR #204, opened 2021-03-05
Fix/radio default color

Christos-Zacharopoulos:
## Description

In this PR we fix the radio input default color and add the missing color property.

[ticket](https://orfium.atlassian.net/browse/DS-36)

## Screenshot

Default color
Capture

default

Colors
colors

## Test Plan

Radio should:
- should render correctly
- should render correctly with color
- should change to checked on click
- disabled should not change to checked on click

### DIFF
--- a/src/components/Radio/Radio.style.ts
+++ b/src/components/Radio/Radio.style.ts
@@ -2,7 +2,6 @@ import { css, SerializedStyles } from '@emotion/core';
 import { rem } from 'polished';
 import { Props } from './Radio';
 import { Theme } from '../../theme';
-import { colorShades, flatColors } from '../../theme/palette';
 
 const hoverColor = 'rgba(0, 0, 0, 0.05)';
 
@@ -85,11 +84,7 @@ export const customRadioStyles = (props: Pick<Props, 'checked' | 'disabled' | 'f
   `;
 };
 
-export const wrapperStyles = (
-  disabled: boolean,
-  fill: typeof flatColors[number],
-  fillShade: typeof colorShades[number]
-) => (theme: Theme): SerializedStyles => css`
+export const wrapperStyles = (disabled: boolean) => (theme: Theme): SerializedStyles => css`
   position: relative;
 
   border-radius: 50%;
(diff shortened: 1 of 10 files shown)


src/components/Radio/Radio.style.ts
```diff
@@ -47,53 +48,56 @@ export const customRadioWrapperStyles = (
   ${focused && !disabled && focusedRadio};
 `;
 
+const determineBoxShadow = ({
+  checked,
+  disabled,
+  filled,
+}: Pick<Props, 'checked' | 'disabled' | 'filled'>) => (theme: Theme) => {
+  if (disabled && !checked) {
+    return `inset 0px 0px 0px ${rem(filled ? '12px' : '2px')} ${theme.utils.getColor(
```
michalisKout, line 56: you can assign `inset 0px 0px 0px ${rem(...)}` in a variable since it used multiple times.

example
```
const getShadowDimensions = (customProp: number) => `inset 0px 0px 0px ${rem(customProp)}`;

.... 
`${getShadowDimensions(12)} ...`
```

src/components/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should render correctly with color', () => {
+    const { container } = render(<Radio color={'teal-400'} />);
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should change to checked on click', () => {
+    const mockOnClick = jest.fn();
```
michalisKout, line 13: you can declare `mockOnClick` in the beforeEach() and reuse it in your tests.

src/components/Radio/Radio.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const { getByTestId } = render(
+      <Radio disabled={true} onChange={mockOnClick} dataTestId={'test-disabled'} />
+    );
+
+    const radio = getByTestId('radio-input-test-disabled');
+
+    fireEvent.click(radio.children[0]);
+
+    expect(mockOnClick).toHaveBeenCalledTimes(0);
+    expect((radio.children[0] as HTMLInputElement).checked).toBeFalsy();
```
michalisKout, line 35: I would suggest using a different approach here as picking the first element may not work in the future (e.g. we add a new span element before the input element). 

example:
```
   const { container } = render(
      <Radio disabled={true} onChange={mockOnClick} dataTestId={'test-disabled'} />
    );

    const radio = container.querySelector('input[type="radio"]') as HTMLInputElement;

    fireEvent.click(radio);

    expect(radio.checked).toBeFalsy();
```

this ensures that we always pick and test the input element.

michalisKout (approved):

PR #219, opened 2021-03-17
[DS-93][DS-95][DS-97] Feat/button transparent

panvourtsis:
## Description

This PR introduce a new design addon for buttons which is the transparent state. This state actually is like the non-filled without the border.

Introduce feature is also applied to icon buttons.

This feature applied already to modal close button and pagination buttons which previously was using the filled version of it.

## Screenshot

Screenshot 2021-03-16 at 2 09 43 PM
Screenshot 2021-03-16 at 2 07 08 PM
Screenshot 2021-03-16 at 2 06 59 PM

## Test Plan
Updated snapshots


Christos-Zacharopoulos (approved):
🎉

PR #220, opened 2021-03-17
Feat/top nav search

Christos-Zacharopoulos:
## Description

In this PR we introduce a custom search component for the Top Navigation component. 

Also updated stories to better showcase the component.

Designs [here](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System?node-id=4458%3A4038)
Ticket [here](https://orfium.atlassian.net/browse/DS-94)

##  Screenshot
Dark
Capture
Light
Capture

## Test Plan

Snapshots for dark and light theme and menu click check.

### DIFF
--- a/src/components/Chart/BarChart/components/CustomTooltipContent/CustomTooltipContent.tsx
+++ b/src/components/Chart/BarChart/components/CustomTooltipContent/CustomTooltipContent.tsx
@@ -1,3 +1,4 @@
+/** @jsxRuntime classic */
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
 import React from 'react';
--- a/src/components/TopAppBar/TopAppBar.stories.mdx
+++ b/src/components/TopAppBar/TopAppBar.stories.mdx
@@ -28,7 +28,7 @@ Regular TopAppBar
 <Preview>
   <Story name="Top App Bar">
     <div style={{padding: 10}}>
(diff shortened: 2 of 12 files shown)


panvourtsis (approved):

michalisKout (commented):
LGTM 👍🏻 . 

_Note:_ We can revisit the component's logic in the future. For example, the `isSearchDisabled` prop is useless in case we do not pass the `onSearchHandler` prop. Having two strongly coupled props may create confusion.

A possible solution to this would be to change the onSearchHandler and add an internal state to Search.

example:
```
- Search.tsx
const [disabled, setDisabled] = setState(false);
...
<div aria-disabled={disabled}>
<input onChange={(e) => { onSearchHandler(e, setDisabled) } } />
<div />
```

Christos-Zacharopoulos:
> LGTM 👍🏻 .
> 
> _Note:_ We can revisit the component's logic in the future. For example, the `isSearchDisabled` prop is useless in case we do not pass the `onSearchHandler` prop. Having two strongly coupled props may create confusion.
> 
> A possible solution to this would be to change the onSearchHandler and add an internal state to Search.
> 
> example:
> 
> ```
> - Search.tsx
> const [disabled, setDisabled] = setState(false);
> ...
> <div aria-disabled={disabled}>
> <input onChange={(e) => { onSearchHandler(e, setDisabled) } } />
> <div />
> ```

@michalisKout 
I agree but this needs to be revisited from a Design perspective. We need to investigate if we want the search to be hidden or disabled or if both are needed.

michalisKout (approved):

PR #231, opened 2021-03-31
fix: apply design bug fixes [DS-110]

whosyourtaco:
## Description

[DS-110]

Fixes color swatches bug in Drawer MenuItems.

Replace Lock and Info icon with their updated counterparts.

Fixes color bug in IconButton props inside Pagination.

Adds color prop in avatar object in UserMenu component. Updated stories to showcase (Menu).

[DS-110]: https://orfium.atlassian.net/browse/DS-110


src/components/Pagination/Pagination.tsx
```diff
@@ -76,7 +75,7 @@ const Pagination = ({
       </div>
 
       <IconButton
-        iconColor={'darkGray'}
+        color={'neutralBlack-700'}
```
kboutsika, line 78: optionally you can avoid curly brakets here.
just `color="neutralBlack-700"`

kboutsika (approved):

src/components/Pagination/Pagination.tsx
whosyourtaco: Sure! 😊

panvourtsis (approved):
LGTM 🎊

PR #235, opened 2021-04-06
Feat/matching module sub components

michalisKout:
## Description
Exposes two components from Asset matching: 
- SectionHeader 
- ActionsToolbox

Added two stories extra in storybook.

We can reuse these components for handling bulk actions in any project.

## Screenshot

In the following image, you can check that the bulk actions section and the header section of the AssetMatching are identical. Therefore we can directly use the SectionHeader component to handle bulk actions

Screenshot 2021-04-06 at 2 56 53 PM

## Test Plan

update snapshots


src/components/AssetMatching/components/ActionsToolbox/SecondaryActions.tsx
```diff
@@ -8,17 +8,22 @@ import ClickAwayListener from 'components/utils/ClickAwayListener';
 
 interface Props {
   secondaryActions: MatchingAction[];
+  actionButtonFill?: boolean;
```
panvourtsis, line 11: can we use our definition of naming bool? https://orfium.atlassian.net/wiki/spaces/ENG/pages/1335591044/General+Code+Rules

src/components/AssetMatching/components/SectionHeader/SectionHeader.tsx
```diff
@@ -14,21 +14,33 @@ interface Props {
   score?: string | number;
   matchingActions?: MatchingAction[];
   styleType: formFieldStyles;
+  actionButtonFill?: boolean;
+  customCheckboxContent?: JSX.Element;
```
panvourtsis, line 18: as this is JSX i would expect capital case

panvourtsis (changes requested):

src/components/AssetMatching/components/SectionHeader/SectionHeader.tsx
michalisKout: Should we use capital case in variables that hold jsx element? 
I thought only functional components should start with capital letter.

for example: 
```
we use button like this:
<Button iconLeft={...} />

iconLeft holds a jsx element.
```

panvourtsis: sorry my fault you are right only for components

panvourtsis (approved):

PR #236, opened 2021-04-06
fix: passes setChecked state handler as an argument in onCheck callback

michalisKout:
## Description

I forgot to expose the setChecked state handler of the checkbox.  🤭

This is a minor fix that passes the state handler to the onCheck callback.


michalisKout:
**UPDATE:** I changed the handling of the checkbox's internal state in a simpler way.

panvourtsis (approved):

PR #237, opened 2021-04-07
fix: adds state update on isChecked param change 

michalisKout:
## Description
- adds a new button in a showcase for checking the component from a parent component.
- adds useEffect and passed isChecked to Checkbox component.

## Screenshot

asset-matching-check-from-outside

## Test Plan

updates snapshot

### DIFF
--- a/src/components/AssetMatching/__snapshots__/AssetMatching.stories.storyshot
+++ b/src/components/AssetMatching/__snapshots__/AssetMatching.stories.storyshot
@@ -1000,7 +1000,7 @@ exports[`Storyshots Design System/Asset Matching SectionHeader 1`] = `
       ⬇️Header section can be used separately for handling bulk actions⬇️
     </h2>
     <header
-      className="css-1kw7jcc-SectionHeader"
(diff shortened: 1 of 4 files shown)


kostasdano:
**[more like a Nice-To-Have, judging from the ACs in the description]** 
even though I can check the component from the parent and then uncheck, it seems that I cannot only uncheck it (that means: check the component directly and then try to uncheck from the parent). I have to click twice to uncheck because the parent will set it to `true` again and then to `false` (no 100% two-way communication between parent-child)

kostasdano (approved):

michalisKout:
> **[more like a Nice-To-Have, judging from the ACs in the description]**
> even though I can check the component from the parent and then uncheck, it seems that I cannot only uncheck it (that means: check the component directly and then try to uncheck from the parent). I have to click twice to uncheck because the parent will set it to `true` again and then to `false` (no 100% two-way communication between parent-child)

you are right @kostasdano , updating it now. Thanks!

PR #238, opened 2021-04-07
fix: removes proposed solution after last PR 

michalisKout:
## Description

Removes proposed solution after the last PR and adds a key that changes dynamically to force rerender the component.

We can use a key that changes dynamically every time the parent components change the boolean prop `isChecked`. This way, we force the section component that contains the checkbox to rerender and pass down the updated value of `isChecked` prop without using the useEffect hook.

## Screenshot

ezgif com-gif-maker


kostasdano (approved):
nice! 🔥

PR #242, opened 2021-04-12
Feat/Optional isActive

Christos-Zacharopoulos:
## Description

On initial render , the active classname is not appended.  This PR introduces the [isActive](https://reactrouter.com/web/api/NavLink/isactive-func) function as an optional param to help solve this issue. 

[Ticket](https://orfium.atlassian.net/browse/DS-125)

### DIFF
--- a/src/components/Drawer/types.ts
+++ b/src/components/Drawer/types.ts
@@ -1,4 +1,6 @@
 import { AcceptedIconNames } from 'components/Icon/types';
+import * as H from 'history';
+import { match } from 'react-router';
 
 export type MenuItem = {
   name: string;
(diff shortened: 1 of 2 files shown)


src/components/Drawer/types.ts
```diff
@@ -1,10 +1,16 @@
...
+import * as H from 'history';
+import { match } from 'react-router';
 
 export type MenuItem = {
   name: string;
   url: string;
   state?: Record<string, any> | null;
   visible: boolean;
   iconName: AcceptedIconNames;
+  isActive?<Params extends { [K in keyof Params]?: string }>(
```
michalisKout, line 11: maybe you should also expose a generic type for LocationState and partially import the Location and LocationState types. This will help us because we have an optional state prop in the MenuItem.

example
```
import { Location, LocationState } from 'history';


isActive?:<Params extends { [K in keyof Params]?: string }, S = LocationState>(
    match: match<Params> | null,
    location: Location<S>
  ): boolean;
```

elektracodes (approved):

PR #249, opened 2021-04-15
fix(asset-matching): adds prop for explicitly handling checkbox

michalisKout:
## Description

- adds `isCheckboxEnabled` prop for explicitly handling checkbox `disabled` prop
- changes the behavior of default probability score element
- updates `SectionHeader` showcase.


panvourtsis (approved):
🎊

PR #252, opened 2021-04-19
[DS-104][URGENT] fix(colors): update all color swatch and add 4 new

panvourtsis:
## Description

This PR updates all swatches that needed update from the palette of DS in figma. It also introduce 4 new ones


elektracodes (approved):

Christos-Zacharopoulos (approved):

Christos-Zacharopoulos:
Color naming is getting a bit strange? Cool gray and light cool gray. It might just be me but lets think if we can change maybe some color names in the next version.

panvourtsis:
> Color naming is getting a bit strange? Cool gray and light cool gray. It might just be me but lets think if we can change maybe some color names in the next version.

why is that? These colors are relatives, is not that weird to me

Christos-Zacharopoulos:
> > Color naming is getting a bit strange? Cool gray and light cool gray. It might just be me but lets think if we can change maybe some color names in the next version.
> 
> why is that? These colors are relatives, is not that weird to me

First of all, do we really need 8 shades of cool gray? 😏 

Second, it starts to get a bit inconsistent. We have `coolGray`  and `lightCoolGray`, `blue` and `lightBlue`. But we don't have the same variants for the rest of the colors. As a user of the DS I would expect to also find a `darkRed`, `darkWarmGray` and even a `gray` which we don't have. 

I believe we need to be more consistent, especially with primitives like palette and colors.

panvourtsis:
> > > Color naming is getting a bit strange? Cool gray and light cool gray. It might just be me but lets think if we can change maybe some color names in the next version.
> > 
> > 
> > why is that? These colors are relatives, is not that weird to me
> 
> First of all, do we really need 8 shades of cool gray? 😏
> 
> Second, it starts to get a bit inconsistent. We have `coolGray` and `lightCoolGray`, `blue` and `lightBlue`. But we don't have the same variants for the rest of the colors. As a user of the DS I would expect to also find a `darkRed`, `darkWarmGray` and even a `gray` which we don't have.
> 
> I believe we need to be more consistent, especially with primitives like palette and colors.

again this is just naming not types or anything similar that you have to bring everything in in the same way. Even material is using these types of naming https://material-ui.com/customization/color/

We can discuss this at the second time but i am not positive to have all those variant names under the same pattern.

PR #253, opened 2021-04-21
Fix/textfield wrapper

Christos-Zacharopoulos:
## Description

In TextFieldWrapper utils component there was a calc on width that was breaking all the textfields. Since we are using everywhere else the value `fill-available` I replaced the calc with it to resolve the problem.

[Ticket](https://orfium.atlassian.net/browse/DS-131)

## Screenshot

Visual proof of the error while using them:
Capture (2)

## Test Plan

Updated snapshots


SVonKarajann (approved):
LGTM

PR #255, opened 2021-04-21
refactor: change all TopAppBar to TopNavBar

kostasdano:
## Description

[DS-130](https://orfium.atlassian.net/browse/DS-130)


michalisKout:
Hmm, I think `TopNavBar` is not semantically correct since we do not include any `nav` HTML element. Our main navigation component is our `Drawer` component.
We also included the `aria-label="Top Application Banner"`.

@panvourtsis any thoughts?

panvourtsis:
> Hmm, I think `TopNavBar` is not semantically correct since we do not include any `nav` HTML element. Our main navigation component is our `Drawer` component.
> We also included the `aria-label="Top Application Banner"`.
> 
> @panvourtsis any thoughts?

In that part you are right but we use two names at the moment. The best thing is to use the correct one and have a new version for it. In that version though we can introduce other things as well so we go back to `next` which will include this +

* deprecation of branded colors
* update to new emotion version

michalisKout:
> > Hmm, I think `TopNavBar` is not semantically correct since we do not include any `nav` HTML element. Our main navigation component is our `Drawer` component.
> > We also included the `aria-label="Top Application Banner"`.
> > @panvourtsis any thoughts?
> 
> In that part you are right but we use two names at the moment. The best thing is to use the correct one and have a new version for it. In that version though we can introduce other things as well so we go back to `next` which will include this +
> 
> * deprecation of branded colors
> * update to new emotion version

Totally agree with you!
So, as an outcome should we keep `TopAppBar` for `master` and move this PR to `next`?
We also should find a more descriptive and accurate name.

elektracodes (dismissed):

elektracodes:
@michalisKout @panvourtsis Have we made a decision on how we would merge this one?

panvourtsis:
> @michalisKout @panvourtsis Have we made a decision on how we would merge this one?

please proceed with the change but before you do make a github ticket for future change

panvourtsis:
I am closing this due to the fact that we will make this to the `next` release

PR #259, opened 2021-04-28
Feat/breadcrumb make data props optional

Christos-Zacharopoulos:
## Description

In this PR we make the data prop passed to breadcrumb optional in order to allow it to work with children without errors.


SVonKarajann (approved):
LGTM

PR #260, opened 2021-05-05
Feat/outlined notifications

Christos-Zacharopoulos:
## Description

This PR introduces the styleType Prop on the Notifications Components.  These are InlineNotification, Banner, Toast and Snackbar.

The styleProp is optional with default value `'elevated'` and takes as values `'elevated'` or `'outlined'`

[design](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System?node-id=201%3A5004)
[ticket](https://orfium.atlassian.net/secure/RapidBoard.jspa?rapidView=40&projectKey=DS&modal=detail&selectedIssue=DS-115)

## Screenshot

banner
inline
snackbar
toast

## Test Plan

Updated snapshots

### DIFF
--- a/src/components/Notification/Banner/Banner.tsx
+++ b/src/components/Notification/Banner/Banner.tsx
@@ -16,7 +16,7 @@ export type Props = {
   /** The type of the Notification */
   type: NotificationTypes;
   /** The style type of the Notification. Defaults to elevated */
-  styleType: NotificationStyleType;
+  styleType?: NotificationStyleType;
   /** The primary call-to-action label of the Notification */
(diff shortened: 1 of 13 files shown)


SVonKarajann (approved):
Try to use theme.spacing when possible. LGTM

Christos-Zacharopoulos:
@SVonKarajann where do you mean? I replaced some usages of  `spacing`   because it was used in `border` and `border-radius` 

If in the future we decide to change the spacing of ictinus, we don't necessarily want to change these values.

PR #264, opened 2021-05-07
Feat/packages upgrade

panvourtsis:
## Description

**WARNING**: many files were changed because of the jsx pragma removal and snapshot updates so keep calm.

This PR introduce two new major things

1) fixes the issues we had with testing and snapshots styling. Now each snapshot shows the class as a new style giving us the opportunity to debug and check if any prop changed easily

2) We also upgraded the follow modules
* emotion 11 🎊 
* latest babel
* latest testing-library
* latest typescript 👀 

NOTE: this will have breaking changes for projects without emotion 11 so for that we introduce this as a `next` release


panvourtsis:
@Orfium/front-end Kind Reminder !
Alt Text

Christos-Zacharopoulos (approved):

PR #267, opened 2021-05-13
feat(table): test ids

kostasdano:
## Description

In this PR test-ids for all `th` and `td` table elements are added:

- `th` : test-id based on the column name (e.g.  column name `First Name` -> `table_header_first_name`)
- `td` : generic test-id with 2d indexing (row and column, e.g. `table_row_4_cell_2` )

Also the user can enter an optional dataTestIdPrefix (e.g. `cmo_table_header_first_name` or `cmo_table_row_4_cell_2`)

### DIFF
--- a/src/components/Table/components/TableCell/TableCell.tsx
+++ b/src/components/Table/components/TableCell/TableCell.tsx
@@ -36,12 +36,15 @@ const TableCell: React.FC<Props> = React.memo(
 
     const tableCellTestId = children
       ? component === 'th' && typeof children === 'string'
-        ? (dataTestIdPrefix ? dataTestIdPrefix + '_table_header_' : 'table_header_') +
-          children.split(' ').join('_').toLowerCase()
-        : (dataTestIdPrefix ? dataTestIdPrefix + '_row_' : 'table_row_') +
-          rowIndex +
-          '_cell_' +
-          index
+        ? (dataTestIdPrefix ? dataTestIdPrefix + '_' : '') +
+          'table_header_' +
+          children
+            .split(' ')
+            .join('_')
+            .toLowerCase()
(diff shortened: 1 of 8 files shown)


src/components/Table/components/TableCell/TableCell.tsx
```diff
@@ -24,10 +27,23 @@ const TableCell: React.FC<Props> = React.memo(
     children,
     type = 'normal',
     padded = false,
+    dataTestIdPrefix,
+    rowIndex,
```
SVonKarajann, line 31: Not sure if a problem, but since both rowIndex and index can be undefined, this can lead to testIds such as 'table_row_undefined_cell_undefined'
table-pr

src/components/Table/components/TableCell/TableCell.tsx
```diff
@@ -24,10 +27,23 @@ const TableCell: React.FC<Props> = React.memo(
...
+    index,
   }) => {
     const theme = useTheme();
     const Component = component;
 
+    const tableCellTestId = children
+      ? component === 'th' && typeof children === 'string'
```
SVonKarajann, line 38: we wrap the checkboxes in th, leading to inconsistencies regarding the rule you are trying to apply 
- th : **table_header_**
- td : **table_row_**


table-th

kostasdano: after a discussion on this, we will keep it as is it

src/components/Table/components/TableCell/TableCell.tsx
kostasdano: resolved

whosyourtaco (approved):
LGTM! 🎉

PR #268, opened 2021-05-14
feat: add filter component

kboutsika:
## Description

implement the a simple filter component based on the design system (no search, no multi selection)

## Screenshot

Στιγμιότυπο 2021-05-14, 2 59 25 μμ
Στιγμιότυπο 2021-05-14, 2 59 17 μμ

### DIFF
--- a/src/components/Filter/Filter.stories.mdx
+++ b/src/components/Filter/Filter.stories.mdx
@@ -40,10 +40,10 @@ A filter with different with style types and a selection handler to return selec
 <Preview>
   <Story name="Filter with style types">
     <Stack>
-       <FilterShowcase styleType="filled" />
-       <FilterShowcase styleType="outlined" />
-       <FilterShowcase styleType="elevated" />
-       <FilterShowcase styleType="transparent" />
+       <FilterShowcase styleType="filled" color="lightCoolGray-400" />
+       <FilterShowcase styleType="outlined" color="lightCoolGray-400" />
+       <FilterShowcase styleType="elevated" color="lightCoolGray-400" />
+       <FilterShowcase styleType="transparent" color="lightCoolGray-400" />
     </Stack>
   </Story>
 </Preview>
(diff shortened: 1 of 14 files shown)


src/components/Filter/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  buttonType,
+  styleType,
+}: BackgroundColorProps) => {
+  if (open) {
+    return defineBackgroundColor(theme, activeCalculatedColor);
+  } else if (hasSelectedValue || styleType === 'transparent') {
+    return 'transparent';
+  } else if (styleType === 'filled' || styleType === 'outlined') {
+    return defineBackgroundColor(theme, calculatedColor, buttonType);
+  } else {
```
panvourtsis, line 22: you don't need an `else` here as a simple return would do the trick

src/components/Filter/Filter.stories.mdx
```diff
@@ -0,0 +1,52 @@
...
+# Filter
+
+A universal Filter component that is a clickable element that can hold many forms.
+
+# Usage
+
+```js
+import { Filter } from '@orfium/ictinus';
+
+<Filter onClick={() => {}}>Click Me</Filter>;
```
panvourtsis, line 17: this usage example is not quite right based on what i saw of your example. Please use a minimal example

src/components/Filter/Filter.style.ts
```diff
@@ -0,0 +1,131 @@
...
+`;
+
+export const buttonSpanStyle = () => () => {
+  return {
+    display: 'flex',
+    alignItems: 'center',
+  };
+};
+
+export const optionsStyle = () => (theme: Theme) => css`
```
panvourtsis, line 21: this is being used on the `menu` component. As this is used in 2 places now can we make a simple one to handle both cases?

src/components/Filter/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  /** Defines the style type of the button */
+  styleType: StyleType;
+};
+
+export type TestProps = {
+  dataTestId?: string;
+};
+
+
+export type BackgroundColorProps = {
```
panvourtsis, line 38: as you use often the following

```
open: boolean;
  theme: Theme;
  hasSelectedValue: boolean;
  calculatedColor: ColorShapeFromComponent;
  activeCalculatedColor: ColorShapeFromComponent;
```
you can make a base type and use this as extension

panvourtsis (changes requested):

src/components/Filter/Filter.stories.mdx
kboutsika: yes you are right i changed it

panvourtsis:
forgot to mention that we need one playground story for this with all the properties in knobs

panvourtsis:
other than that this PR LGTM 🎊

panvourtsis (approved):

PR #271, opened 2021-05-18
Feat/import order

Christos-Zacharopoulos:
## Description

DON'T FREAK OUT! A lot of files were changed since this changed the order of import in almost all the `.ts` and `.tsx` files.

In this PR we add the `import/order` rule to eslintrc for specific type of import order. 

I created a first draft according to what we had discussed in a meeting. Feel free to provide feedback and propose changes.


src/test/utils.tsx
```diff
@@ -1,7 +1,7 @@
-import React from 'react';
 import { render, screen } from '@testing-library/react';
 import { renderHook } from '@testing-library/react-hooks';
 import userEvent from '@testing-library/user-event';
+import React from 'react';
```
whosyourtaco, line 4: It looks like the React imports don't get sorted the way we want (always on top)

whosyourtaco (commented):
Great job! This will help us a lot. 

I just have two comments to make. The first is about the 'react' imports which, as you can see in my other comment, look like they don't get auto-sorted by eslint.

The second one is that IMO the import list looks cluttered without a newline in-between groups.

With that in mind, I'm proposing this configuration:

```
"import/order": [
      "error",
      {
        "groups": ["builtin", "external", "internal"],
        "pathGroups": [
          {
            "pattern": "react",
            "group": "external",
            "position": "before"
          }
        ],
        "pathGroupsExcludedImportTypes": ["react"],
        "alphabetize": {
          "order": "asc",
          "caseInsensitive": true
        },
        "newlines-between": "always"
      }
    ]
```

@Orfium/front-end thoughts?

mkarajohn (approved):

Christos-Zacharopoulos:
@whosyourtaco 

I had tried the `newlines-between` option but I was not sure if we wanted it or not since the imports are automatically handled by the IDE. Nevertheless, I will add it since it's much cleaner.

For the React import. I had it initially like this but, [react 17](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) doesn't require imports to use JSX so in many files it won't even be there and when it is we will be importing hooks or types.

whosyourtaco (approved):
LGTM! 🚀

PR #276, opened 2021-05-27
fix(select): on focus stroke issue

kostasdano:
## Description

[DS-101](https://orfium.atlassian.net/browse/DS-101)

When the Select component is onFocus state the stroke style is not applied always. The stroke appears only if you click on the input field; if you click the downArrow icon the stroke is not applied. With this PR the stroke is applied in both cases (input field click or downArrow icon click)

### Before

before

### After

after

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -177,9 +177,9 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps>(
             max-width: ${rem(620)};
             & > div:nth-of-type(1) > div {
               ${open &&
-                status != 'error' &&
+                status !== 'error' &&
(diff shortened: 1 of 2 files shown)


mkarajohn (approved):

src/components/Select/Select.tsx
```diff
@@ -174,9 +175,16 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps>(
             position: relative;
             min-width: ${rem(150)};
             max-width: ${rem(620)};
+            & > div:nth-of-type(1) > div {
+              ${open &&
+                status != 'error' &&
```
panvourtsis, line 180: != > === only for conventional reasons in this case

panvourtsis (approved):

PR #277, opened 2021-05-27
chore: qa design

kboutsika:
## Description
[DS-144](https://orfium.atlassian.net/browse/DS-144)

+ Design oversight in elevated style: “Visited” State should also have elevation. The only state in Elevated style that does not have elevation is focused. 

+ add a max hight for menu (253px)

+ Default option should always be the first list item and font should have medium weight - design oversight

+ If user has selected any other option than default it should be visible by opening menu

### DIFF
--- a/src/components/Filter/Filter.style.ts
+++ b/src/components/Filter/Filter.style.ts
@@ -50,7 +50,7 @@ export const buttonStyle = ({
       activeCalculatedColor,
       calculatedColor,
     }),
-    boxShadow: styleType === 'elevated' && !hasSelectedValue ? theme.elevation['01'] : 'none',
+    boxShadow: styleType === 'elevated' ? theme.elevation['01'] : 'none',
 
     border: getBorder({
       styleType,
@@ -66,7 +66,7 @@ export const buttonStyle = ({
       border: hasSelectedValue
         ? `solid 1px ${stateBackgroundColor(theme, 'hover', calculatedColor, true)}`
         :  styleType === 'outlined'? `solid 1px ${stateBackgroundColor(theme, 'hover', calculatedColor, true)}` : 'none',
-      boxShadow: styleType === 'elevated' && !hasSelectedValue ? theme.elevation['02'] : 'none',
+      boxShadow: styleType === 'elevated' ? theme.elevation['02'] : 'none',
     },
     ':active': {
       backgroundColor: !disabled && !open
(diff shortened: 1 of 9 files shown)


src/components/Filter/Filter.style.ts
```diff
@@ -50,7 +50,7 @@ export const buttonStyle = ({
       activeCalculatedColor,
       calculatedColor,
     }),
-    boxShadow: styleType === 'elevated' && !hasSelectedValue ? theme.elevation['01'] : 'none',
+    boxShadow: styleType === 'elevated' ? theme.elevation['01'] : 'none',
```
Christos-Zacharopoulos, line 53: For consistency reasons as well as readability I would suggest to follow a similar pattern to `wrapperStyleSwitch` that's defined in `src/components/AssetMatching/AssetMatching.style.ts`  when for properties based on styleType.

src/components/utils/DropdownOptions/index.tsx
```diff
@@ -10,7 +10,7 @@ export type MenuOptions = {
 };
 
 export const optionsStyle = ({ menuPosition }: MenuOptions) => (theme: Theme) => css`
-  max-height: 400px;
+  max-height: 253px;
```
Christos-Zacharopoulos, line 13: Use `rem(253)` here.

src/components/Filter/Filter.tsx
```diff
@@ -76,12 +78,31 @@ const Filter: React.FC<Props & TestProps> = props => {
         </button>
         {open && (
           <div css={optionsStyle({})(theme)} data-testid="filter-menu">
+            <button
+              css={{
```
Christos-Zacharopoulos, line 82: For consistency reasons, since we are using functions to pass style to the css prop, move this to the `Filter.styles` file.

src/components/Filter/Filter.tsx
```diff
@@ -76,12 +78,31 @@ const Filter: React.FC<Props & TestProps> = props => {
...
+            </button>
             {items &&
-              items.map((option, index) => (
+              items
+                .filter(option => option.value !== defaultValue.value) //filter options just in case the default value is included
+                .map((option, index) => (
                 <button
                   css={{
-                    backgroundColor: theme.palette.white,
+                    backgroundColor: option.value === selectedItem?.value ? darken(0.05, theme.palette.white) : theme.palette.white,
```
Christos-Zacharopoulos, line 103: Same here.

Christos-Zacharopoulos (changes requested):

src/components/Filter/Filter.style.ts
kboutsika: Excuse my delayed reply! Yes that is an absolutely nice suggestion! thank you! I really tried to simplify the css of this component but didn't manage to provide a "clean" solution. The style of the wrapper is not relevant only with the `styleType` but also if the the menu is open and if there is and selectedItem, so a switch case won't change a lot these parts. However if you have any specific suggestions on that i would be more than grateful if you share with me.

src/components/Filter/Filter.tsx
kboutsika: yes i changed thank you!

src/components/Filter/Filter.style.ts
Christos-Zacharopoulos: Well , I was only talking about the boxShadow prop that was visible in the diff. After checking out more the code, I can see what you are talking about.

Christos-Zacharopoulos (approved):

PR #286, opened 2021-06-07
[DS-176]: Feat/ds 176

michalisKout:
## Description

- fixes loader positioning in dom

[jira task](https://orfium.atlassian.net/browse/DS-176)

### before fix
before_fix

### after fix
after_fix

## Test Plan

- updates snapshots

### DIFF
--- a/src/components/Loader/components/DotsLoader/DotsLoader.style.tsx
+++ b/src/components/Loader/components/DotsLoader/DotsLoader.style.tsx
@@ -1,12 +1,13 @@
 import { css, SerializedStyles } from '@emotion/core';
 
 import { Theme } from 'theme';
+import { rem } from 'polished';
 
 export const getDotsLayout = (delay: number, animation: string, theme: Theme, left?: number) => css`
   left: ${left && `${left}px`};
-  width: 6px;
-  height: 6px;
-  border-radius: 5px;
+  width: ${rem(6)};
+  height: ${rem(6)};
+  border-radius: ${rem(5)};
   background-color: ${theme.utils.getColor('lightGray', 300)};
   color: ${theme.utils.getColor('lightGray', 300)};
   animation: ${animation};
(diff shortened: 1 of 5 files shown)


src/components/Loader/components/DotsLoader/DotsLoader.style.tsx
```diff
@@ -13,8 +13,15 @@ export const getDotsLayout = (delay: number, animation: string, theme: Theme, le
   animation-delay: ${delay}s;
 `;
 
-export const dotsContainer = () => (theme: Theme): SerializedStyles => css`
+export const dotsWrapper = css`
+  width: 26px;
```
panvourtsis, line 17: can we use rem here as this will possible break in responsive width

src/components/Loader/components/DotsLoader/DotsLoader.style.tsx
```diff
@@ -13,8 +13,15 @@ export const getDotsLayout = (delay: number, animation: string, theme: Theme, le
...
   position: relative;
+  height: 6px;
+`;
+
+export const dotsContainer = () => (theme: Theme): SerializedStyles => css`
+  position: absolute;
+  left: 10px;
```
panvourtsis, line 24: same here

panvourtsis (changes requested):

src/components/Loader/components/DotsLoader/DotsLoader.style.tsx
michalisKout: Since the loader's style depends on pixels I didn't want to change the units to relative just for the fix.

Also I am not sure if we need to include relative units for components that behave like loaders.

Any thoughts on this?

panvourtsis: hmm i see your point since this is small and will always depond in this but again if we use a `meta` tag to define the viewport and device width this will be very small

panvourtsis:
@michalisKout reminder

michalisKout:
@panvourtsis reminder for approval

panvourtsis (approved):

PR #296, opened 2021-06-11
[No-Jira] Feat/select area width fix

Christos-Zacharopoulos:
## Description

This PR resolves a minor issue on DS on sizing of outlined TextField.

The outlined TextField has it's border visually created by a box-shadow. This results that the width of the component is larger than the width that the DOM gives to the component.

No ticket exists for this. If we consider it useful , I will create one to pass this for Design QA.

## Screenshot

Before the proposed change
before

with the proposed change
after

## Test Plan

Updated snapshots

### DIFF
--- a/src/components/DatePicker/__snapshots__/DatePicker.stories.storyshot
+++ b/src/components/DatePicker/__snapshots__/DatePicker.stories.storyshot
@@ -3475,7 +3475,7 @@ exports[`Storyshots Design System/DatePicker DatePicker Range 1`] = `
               className="css-o66dxn-TextInputWrapper"
             >
               <div
-                className="css-71prqf-TextInputWrapper"
+                className="css-1wqyga6-TextInputWrapper"
               >
                 <div
                   className="css-19bjb4x-TextInputWrapper"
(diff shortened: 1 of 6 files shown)


panvourtsis:
The solution seems a bit laggy isn't it?

**New one**
ezgif com-gif-maker (5)

**Old one**
ezgif com-gif-maker (6)

Christos-Zacharopoulos:
> The solution seems a bit laggy isn't it?
> 
> **New one**
> ezgif com-gif-maker (5)
> 
> **Old one**
> ezgif com-gif-maker (6)

Yea it actually seems very laggy watching them like this next to each other.  I'll try to figure out something later since this is not an immediate request. Feel free to pitch any ideas that could help.

panvourtsis:
any news heres?

mkarajohn:
Have you tried making it an inset shadow instead of a border? May help with the perceived lagginess

Christos-Zacharopoulos:
> any news here?

have not had the time to check this again.

Christos-Zacharopoulos:
> Have you tried making it an inset shadow instead of a border? May help with the perceived lagginess

This was the previous implementation I think , but it was messing up the spacings in the design team , since shadow is outside the space a component takes in the DOM.

mkarajohn:
> > Have you tried making it an inset shadow instead of a border? May help with the perceived lagginess
> 
> This was the previous implementation I think , but it was messing up the spacings in the design team , since shadow is outside the space a component takes in the DOM.

From what I can see in the diff, it was not an inset shadow :eyes:

panvourtsis:
closing this for now as there isn't something new

PR #299, opened 2021-06-15
feat: add search to filters

kboutsika:
## Description

Add search functionality to to Filter component.

This MR extends the functionality of the simple Filter component. Extra props added to the component in order to be able to search the options. Search can be
+ just filtering of the existing options
+ an async search. in this case the component trigger a function that is provided as prop when the user types. This is used for live search

[DS-145](https://orfium.atlassian.net/browse/DS-145)

## Screenshot
Στιγμιότυπο 2021-06-16, 9 56 35 πμ


panvourtsis:
@kboutsika can you elaborate of the final outcome on the ticket please? what i expect to see ?

kboutsika:
@panvourtsis yes! sorry I missed the screenshots! I added them alone with the description of the MR

src/components/Filter/Filter.test.tsx
```diff
@@ -1,44 +1,43 @@
 import React from 'react';
 import '@testing-library/jest-dom/extend-expect';
 import userEvent from '@testing-library/user-event';
-
+import { render, screen } from 'test';
 import Filter from './Filter';
-import { render, screen } from '../../test';
+import { waitFor } from '../../test';
```
whosyourtaco, line 6: You could add the `waitFor` import to the `'test'` one :)

src/components/Filter/Filter.tsx
```diff
@@ -1,38 +1,49 @@
...
   buttonStyle,
   childrenWrapperStyle,
   buttonSpanStyle,
   labelSpanStyle,
   wrapperStyle,
+  menuStyle,
 } from './Filter.style';
+import Options from './components/Options/Options';
+import { useMemo } from 'react';
+import { debounce } from 'lodash';
```
whosyourtaco, line 23: You could move these 2 imports higher :)

src/components/Filter/Filter.tsx
```diff
@@ -80,27 +128,23 @@ const Filter: React.FC<Props & TestProps> = props => {
...
-                <button
-                  type="button"
-                  css={optionStyle(option, selectedItem)(theme)}
-                  key={`${option.value}-${index}`}
-                  onClick={() => {handleSelect(option);}}
-                >
-                  {option.label}
-                </button>
-              ))}
+          <div css={menuStyle()(theme)} data-testid="filter-menu">
```
whosyourtaco, line 131: Maybe you could add the `dataTestId` as a prefix here? Can this component be used multiple times in the same page?

src/components/Filter/components/Options/Options.style.ts
```diff
@@ -0,0 +1,35 @@
...
+`;
+
+export const emptyStyle = () => (theme: Theme) => css`
+  color: ${theme.utils.getColor('lightGray', 600)};
+  height: ${rem(48)};
+  padding: 0 ${rem(16)};
+  font-size: ${theme.typography.fontSizes['14']};
+  display: flex;
+  align-items: center;
+  justify-content: center;
```
whosyourtaco, line 33: You could use the `${flexCenter}` function from `src/theme/functions.ts`

elektracodes (approved):

src/components/Filter/Filter.tsx
```diff
@@ -52,6 +63,43 @@ const Filter: React.FC<Props & TestProps> = props => {
     onSelect(option);
   }
 
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
Christos-Zacharopoulos, line 66: @panvourtsis is this supposed to have the exact same functionality as async select component?

If yes, the we could abstract this logic too a separate utils file and reuse it in both components.

src/components/Filter/components/SearchInput/SearchInput.style.ts
```diff
@@ -0,0 +1,29 @@
...
+      outline: none;
+      border: none;
+      box-shadow: none;
+    }
+  }
+`;
+
+export const iconWrapper = () => () => css`
+  display: flex;
+  gap: 25px;
```
Christos-Zacharopoulos, line 28: Minor but let's change this to `rem`

src/components/Filter/components/SearchInput/SearchInput.style.ts
```diff
@@ -0,0 +1,29 @@
+import { css } from '@emotion/core';
+
+import { Theme } from 'theme';
+
+export const textFieldWrapper = () => (theme: Theme) => css`
+  border-bottom: 1px solid ${theme.utils.getColor('lightCoolGray', 200)};
+  
+  div:first-of-type {
```
Christos-Zacharopoulos, line 8: Is there not any variation of the TextField component full filling our needs here? 

It's small overrides that are happening but I would prefer to use things unchanged. This is more of a design question. 

Adding @panvourtsis to the discussion.

src/components/Filter/types.ts
```diff
@@ -29,9 +29,17 @@ export type Props = {
...
   disabled?: boolean;
-};
-
-export type TestProps = {
+  /** if the options are searchable */
+  isSearchable?: boolean;
+  /** if the component is used asynchronously */
+  isAsync?: boolean;
+  /** the function to fetch new options */
+  onAsyncSearch?: (term: string) => void;
```
Christos-Zacharopoulos, line 37: Maybe symblify the props by having a `onSearch` function and a `isAsync` flag.  If on `onSearch`  exists the `isSearchable` would be true so it's overkill to have both.

src/components/Filter/components/Options/Options.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const theme = useTheme();
+
+  return (
+    <div css={optionsStyle()(theme)}>
+      {shouldDisplayDefaultOption && <button
+        type="button"
+
+        data-testid={generateTestDataId('default-option', dataTestId)}
+        css={defaultOptionStyle(defaultValue, selectedItem)(theme)}
+        onClick={() => {onSelect(defaultValue);}}
```
Christos-Zacharopoulos, line 37: OCD kicking in! Could you remove the brackets here or add new lines? 

image

src/components/Filter/components/SearchInput/SearchInput.style.ts
kboutsika: I did not find any prop/way in TextField component to do this without adding custom css.

I will confirm with the design team again and let you know

src/components/Filter/components/Options/Options.tsx
kboutsika: yes you are right thank you

src/components/Filter/types.ts
kboutsika: Thank you for the suggestion. I think is `isSearchable` is required but for sure we can have avoid `isAsync` and keep only `onAsyncSearch`. I will give a try.

When just `isSearchable` is true then the component will filter the options inside so no `onSearch` is needed. But when `onAsyncSearch` is passed then the options are not filtered, the component just triggers `onAsyncSearch`.

src/components/Filter/Filter.tsx
kboutsika: Inside `Select` component we have used `onInput` and not `onChange` but yes the implementation is almost the same. I would rather not to touch other components in this MR. How about to open a ticket to refactor these parts to both components in a separated PR?

src/components/Filter/types.ts
kboutsika: @Christos-Zacharopoulos ok I changed it. please have a look when possible! thanks!

src/components/Filter/Filter.tsx
Christos-Zacharopoulos: The type of the change event could be abstracted as a generic type and be passed by each component so it could satisfy both cases. 
In general I prefer things to be changed on the spot and not add tech debt but I don't mind this happening in a separate branch as long as it is documented and changed.

src/components/Filter/types.ts
Christos-Zacharopoulos: image

src/components/Filter/components/SearchInput/SearchInput.style.ts
kboutsika: @Christos-Zacharopoulos yes just confirmed with the design team...we don' t want here the behavior of the `TextInput`.

src/components/Filter/Filter.tsx
kboutsika: ok yes i agree for the type of change event and i made the change to both files. thank you!

regarding the the re-usable logic, I am happy to do the change, but to be honest i am not sure that we have much to gain. Maybe i miss something but it would a function that should have to take 5 arguments (the isSearchable prop, the setSearchValue function, the event, the minCharactersToSearch prop and the debouncedOnChange function) and its a lot, isn't?

Christos-Zacharopoulos: It might be a lot but I see more value in keeping all the search logic in one place rather than having to maintain the same logic in two (for now) places.

I would expect all the search functionalities in the app to be the same and work with the same way, something that, as we keep adding new components and the code base gets larger,  it's going to be harder to maintain duplicate code, let alone remember it exists. 

I am not going to block the ticket by insisting even more on this. At least add a TODO in both places that the code is pretty much duplicated and we need to figure out a way of refactoring it. ( Also it will help maintain it as long as it stays like this )

kboutsika: @Christos-Zacharopoulos thank you for the nice conversation and suggestion! I think you are right and I made the changes as you said :) 

Thanks again!

PR #303, opened 2021-06-16
[DS-153] Feature/tooltip

panvourtsis:
## Description

This PR introduce a new component for our system, a tooltip. 

This tooltip is only holding plain text. You can define position but based on the screen this will change if there is no space to show it on the pre-defined location.

A new system config name `overrides` was introduced. Its mainly for tooltip ATM in order to define its color and shade for the whole system. The logic is that tooltip will have only one color that will be unique for the whole project. This way we may introduce new overrides in the future.

[Ticket](image)
[Design](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System?node-id=659%3A13)

Note: Asset Matching was changed due to the fact we were using `tippyjs` for the tooltip and now i change it.

## Screenshot

image
image

## Test Plan

* storyshots

### DIFF
--- a/src/components/AssetMatching/components/Asset/AssetExternalLink.tsx
+++ b/src/components/AssetMatching/components/Asset/AssetExternalLink.tsx
@@ -1,29 +1,20 @@
 import React, { FC } from 'react';
-import Tippy from '@tippyjs/react';
 import Styles from './Asset.style';
 import Icon from '../../../Icon';
-import 'tippy.js/dist/tippy.css';
+import Tooltip from 'components/Tooltip';
 
 export interface ExternalLinkProps {
   url: string;
   tooltipContent: string;
 }
 
-// @TODO: Introduce a generic tooltip component based on the ACs of the story: https://orfium.atlassian.net/browse/DS-153
-
 const AssetExternalLink: FC<ExternalLinkProps> = ({ tooltipContent, url }) => {
   return (
-    <Tippy
-      content={tooltipContent}
-      arrow
-      placement={'right'}
-      allowHTML={false}
-      css={Styles.tooltip}
-    >
+    <Tooltip content={tooltipContent} placement={'right'}>
       <a css={Styles.externalLink} href={url} rel="noreferrer" target="_blank">
         <Icon name={'externalLinkV2'} size={10} color={'lightGray'} variant={600} />
       </a>
-    </Tippy>
+    </Tooltip>
   );
 };
 
--- a/src/components/Tooltip/Tooltip.style.ts
+++ b/src/components/Tooltip/Tooltip.style.ts
@@ -2,11 +2,15 @@ import { rem } from 'polished';
 import { Theme } from '../../theme';
 import { css, SerializedStyles } from '@emotion/core';
 import { pickTextColorFromSwatches } from '../../theme/palette';
-import { TooltipSize } from './Tooltip';
+import { TooltipPlacement, TooltipSize } from './Tooltip';
 
-export const tooltipStyle = ({ size }: { size: TooltipSize }) => (
-  theme: Theme
-): SerializedStyles => {
+export const tooltipStyle = ({
+  placement,
+  size,
+}: {
+  placement: TooltipPlacement;
+  size: TooltipSize;
+}) => (theme: Theme): SerializedStyles => {
   const color = theme.overrides.tooltip.background.color;
   const shade = theme.overrides.tooltip.background.shade;
   const backgroundColor = theme.utils.getColor(color, shade);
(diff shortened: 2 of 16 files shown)


whosyourtaco (approved):
LGTM 🎉

whosyourtaco:
Only thing I would comment on is the sizes of the tooltip, because `normal` and `medium` sound kind of the same to me, but they represent different sizes.

src/index.tsx
```diff
@@ -63,5 +66,7 @@ export {
   Palette,
   Typography,
   Elevation,
+  overrides,
```
Christos-Zacharopoulos, line 69: Do we need to expose this?

src/components/AssetMatching/components/Asset/AssetExternalLink.tsx
```diff
@@ -1,29 +1,20 @@
...
 const AssetExternalLink: FC<ExternalLinkProps> = ({ tooltipContent, url }) => {
   return (
-    <Tippy
-      content={tooltipContent}
-      arrow
-      placement={'right'}
-      allowHTML={false}
-      css={Styles.tooltip}
-    >
+    <Tooltip content={tooltipContent} placement={'right'}>
```
michalisKout, line 13: before: 
Screenshot 2021-06-17 at 10 20 29 AM
after:
Screenshot 2021-06-17 at 10 20 39 AM

You should also pass the `Styles.tooltip` to the new tooltip component.

src/index.tsx
panvourtsis: in case we want to know overrides, i didnt had a case in mind but we expose all others for the same reasoning

michalisKout (changes requested):

src/components/Tooltip/Tooltip.style.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (size === 'large') {
+      return theme.typography.fontSizes['16'];
+    } else if (size === 'small') {
+      return theme.typography.fontSizes['12'];
+    }
+
+    return theme.typography.fontSizes['14'];
+  };
+
+  return css`
```
michalisKout, line 24: you can avoid all the `!important` by adding a new parent to the tooltip. Example:
jsx
```
   <div css={tooltipStyle({ size, placement })}>
        <ReactTooltip id={tooltipID} place={placement} effect="solid" className="tooltip">
          {content}
        </ReactTooltip>
      </div>
```

styles
```
   > div {
      color: ${pickTextColorFromSwatches(color, shade)};
      background-color: ${backgroundColor};
      max-width: ${rem(256)};
      padding: ${theme.spacing.sm};
      font-size: ${defineFontSizeBasedOnTooltipSize(size)};
      line-height: 110%;
      border-radius: ${theme.spacing.sm};
      text-align: justify;
    }

    .show {
      opacity: 1 !important;
    }
    ....
```

src/components/Tooltip/Tooltip.style.ts
```diff
@@ -0,0 +1,63 @@
...
+    max-width: ${rem(256)};
+    padding: ${theme.spacing.sm} !important;
+    font-size: ${defineFontSizeBasedOnTooltipSize(size)} !important;
+    line-height: 110%;
+    border-radius: ${theme.spacing.sm};
+    text-align: justify;
+    &.show {
+      opacity: 1 !important;
+    }
+    &.place-right {
```
michalisKout, line 36: you can make this more dynamic

example:
```
const getPlacement = (backgroundColor: string, placement?: string) =>
  placement &&
  css`
    .place-${placement} {
      ::after {
        border-${placement}-color: ${backgroundColor} !important;
      }
    }
  `;
```

usage
${getPlacement(placement, backgroundColor)}

src/components/Tooltip/Tooltip.style.ts
michalisKout: You can also replace  `> div` with `> ${tooltipId}` to make it more secure for future changes 😄

src/components/AssetMatching/components/Asset/AssetExternalLink.tsx
panvourtsis: dont worry about this, its configurable via theme provider. I already talked with elisavet and i would reach to you for the config needed

src/components/Tooltip/Tooltip.style.ts
panvourtsis: i though of this but is it really !important ? (see what i did there?)

michalisKout: I think it's always !important to avoid `!important;`

PR #312, opened 2021-06-23
[DS-134]: Button loading state

michalisKout:
## Description

- adds button loading state handling for async operators
- introduces new spinner loader 
- refactors Loader component to make it more scalable for the upcoming loaders

[jira task](https://orfium.atlassian.net/secure/RapidBoard.jspa?rapidView=40&projectKey=DS&modal=detail&selectedIssue=DS-134)

## Screenshot

ezgif com-gif-maker (2)

Async Buttons => https://ictinus-fix-ds-134-o0590dditrk.herokuapp.com/?path=/story/design-system-button--async-button
Refactored Loaders => https://ictinus-fix-ds-134-o0590dditrk.herokuapp.com/?path=/story/design-system-loaders--available-loaders

## Test Plan

- updates snapshots


michalisKout:
@panvourtsis 
Should we create a new task for the new spinner loader or should I create it in this PR? 
Also should we include the async handling for IconButton too?

panvourtsis:
i am ok for handling it here

michalisKout:
> i am ok for handling it here

ok! Converting the PR to draft to implement the spinner

src/components/Button/Button.tsx
```diff
@@ -67,13 +82,14 @@ const Button: React.FC<Props & TestProps & EventProps> = props => {
         iconRight,
         childrenCount: React.Children.count(children),
       })}
-      onClick={onClick}
+      onClick={handleAsyncOperation}
```
kboutsika, line 85: I am so familiar with the codebase, so maybe i miss something. If there there is a `submit` button inside a form for example, so the form should be submitted with enter key or with submit event, right? If so, how we can set the button in the loading state?

src/components/Button/Button.tsx
```diff
@@ -2,12 +2,20 @@
...
+  buttonSpanStyle,
+  buttonStyle,
+  centralizedLoader,
+  childrenWrapperStyle,
+  iconStyle,
+} from './Button.style';
 import { useTypeColorToColorMatch } from '../../hooks/useTypeColorToColorMatch';
+import { useLoading } from 'hooks/useLoading';
+import Loader from '../Loader';
+import { useRef } from 'react';
```
kboutsika, line 18: there is already everything imported from `react` in line 4
`import * as React from 'react';`. How about to do the below?
+ either `import * as React from 'react';` and `React. useRef<HTMLSpanElement>(null)`, or
+ `import React, { useRef } from 'react';`

src/components/Button/Button.tsx
michalisKout: At the moment, our buttons do not have a specific type, which semantically incorrect. Although, based on the buttons' usage is not required for now.
I already planned to open a bug in the DS board. We can introduce a new prop buttonType ('submit', 'reset', 'button'). 

Then we can investigate accessing the loading state using keys or button types.

kboutsika: ok! thank you for the reply!

kboutsika (approved):

panvourtsis:
reminder !

michalisKout:
> reminder !

waiting for your approval on this :)

Christos-Zacharopoulos:
> > reminder !
> 
> waiting for your approval on this :)

image

src/components/Loader/Loader.stories.mdx
```diff
@@ -1,8 +1,9 @@
 import { Meta, Story, Preview, Props } from '@storybook/addon-docs/blocks';
 
 import Loader from './Loader';
+import {loaderTypes} from './Loader.utils';
 
-<Meta title="Design System/Loader" component={Loader} />
+<Meta title="Design System/Loaders" component={Loader} />
```
panvourtsis, line 6: why plural? In that case Button could be Buttons

src/components/Loader/components/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { FC } from 'react';
+import { spinnerContainer } from './Spinner.style';
+import { generateTestDataId } from '../../../../utils/helpers';
+
+interface Props {
```
panvourtsis, line 5: extend TestId from utils. There is already a type for that

src/components/Loader/Loader.stories.mdx
michalisKout: I thought of this like a group of loaders. Like having different geometric shapes.

Geometric Shapes (plural) => Square, triangle, circle etc...
Loaders => Spinner, dots, indeterminate etc... 

but i can revert it to `Loader` if this is blocking

panvourtsis: no its not blocking it was just confusing as it refers to a single point system like buttons

PR #318, opened 2021-06-30
fix: improve import/order eslint config

whosyourtaco:
Based on the previous work of @Christos-Zacharopoulos here: [Github](https://github.com/Orfium/orfium-ictinus/pull/271), this PR will extend the configuration of the eslint plugin, better encapsulating all kinds of imports.

NOTE: As noted in the library, imports like `import 'tippy.js/dist/tippy.css';` are left untouched because order might be important.

Also, removed some leftover imports that were unnecessary in some test files, since it was already added in the test's `setup.ts` file.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -8,7 +8,12 @@
     "plugin:react-hooks/recommended",
     "plugin:react/recommended"
   ],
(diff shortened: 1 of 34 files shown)


Christos-Zacharopoulos (approved):

panvourtsis:
kudos 🎊

PR #321, opened 2021-07-01
[NO-JIRA] feat(workflow): add ci release for github packages

panvourtsis:
## Description
This PR introduce a new mechanism to release ictinus to github-packages as well. The vision is to have everything under packages here but to not break the Ictinus project. So only ictinus will be on npm registry

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -16,6 +16,14 @@
           "changelogFile": "CHANGELOG.md"
         }
       ],
-      "@semantic-release/npm"
+      [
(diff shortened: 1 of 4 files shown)


.releaserc
```diff
@@ -16,6 +16,14 @@
           "changelogFile": "CHANGELOG.md"
         }
       ],
-      "@semantic-release/npm"
+      [
+        "@amanda-mitchell/semantic-release-npm-multiple",
```
michalisKout, line 20: 🙏🏻

michalisKout (approved):

PR #323, opened 2021-07-05
[DS-195] Chat Bubble Icon

kostasdano:
## Description

[DS-195](https://orfium.atlassian.net/browse/DS-195)


src/components/Icon/assets/chat-bubble.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
michalisKout, line 1: Hello @kostasdano, I suggest to remove the width and the height from the svg since we handle these properties using css.

michalisKout (approved):

src/components/Icon/assets/chat-bubble.svg
kostasdano: nice catch, thanks! :D

PR #324, opened 2021-07-06
feat: [DS-211] add custom component to drawer

kboutsika:
## Description
[DS-211](https://orfium.atlassian.net/browse/DS-211)

+ added an optional render prop over the menu items in case we need to display a custom component 

## Screenshot
Στιγμιότυπο 2021-07-06, 11 14 18 πμ
Στιγμιότυπο 2021-07-06, 11 14 12 πμ


mkarajohn (approved):

mkarajohn:
While the implementation is ok, is very specific to the usecase scenario of the ADR client. In a future update I would suggest the Drawer be separated from its contents (make it a render-props component that passes them down to the function used as `children`) so that the user can use anything as the content of the drawer.

src/components/Drawer/Drawer.tsx
```diff
@@ -14,6 +15,8 @@ export type Props = {
   setExpanded: (v: boolean) => void;
   /** The menu items to be displayed in the drawer */
   menuItems: MenuItem[];
+  /** Render prop function to display something over the Navigation */
+  renderHeader?: () => React.ReactNode;
```
michalisKout, line 19: You can change the type to ```renderHeader?: React.FC```

Christos-Zacharopoulos:
> While the implementation is ok, it is very specific to the usecase scenario of the ADR client. In a future update I would suggest the Drawer be separated from its contents (make it a render-props component that passes them down to the function used as `children`) so that the user can use anything as the content of the drawer.

This is planned for V4 since it would be a breaking change right now.

src/components/Drawer/Drawer.tsx
mkarajohn: That would mean `renderHeader` returns a function. From what I udnderstand this is following the render props pattern, so renderHeader is in fact returning a `ReactNode` or a `ReactElement` (objects) and not a function (`FC`)

michalisKout: render props patterns requires children prop as function to work properly.

Since this is just a JSX.Element prop. @kboutsika You can just add a type of `JSX.Element`or `ReactNode` without `() =>`. Like we did in the Button for example https://github.com/Orfium/orfium-ictinus/blob/master/src/components/Button/Button.tsx#L33.

kboutsika: yes thank you both for the review! @michalisKout this is just another approach that for me its also pretty fine! However it more like personal preference if you you want to use a render prop or a JSX.Element.

PR #325, opened 2021-07-06
fix: add key to content and fix localized content issue [DS-213]

whosyourtaco:
[DS-213]

Fixes issue with localized content as table header where sorting would break, since changing the language would result in a discrepancy between the Table's sorting state and the selected Column's name.

The content key has been replaced by a label/key pair in order to fix this issue and generally be more consistent.
`label` should contain the actual text displayed, be it localized or not.
`key` should be a unique identifier of the column. This value is also used to generate some necessary test-ids.

[DS-213]: https://orfium.atlassian.net/browse/DS-213


src/components/Table/Table.test.tsx
```diff
@@ -97,17 +97,23 @@ describe('Table', () => {
       <Table
         columns={[
           {
-            content: 'Title',
+            content: {
+              label: 'Title',
+              key: 'title',
```
michalisKout, line 102: You can rename the `key` prop to `sortingKey` to indicate its purpose.

src/components/Table/Table.tsx
```diff
@@ -223,20 +223,20 @@ function Table<T>({
...
                       onClick={() => {
                         if (!isItemString(item) && item.isSortable) {
-                          handleSorting(item.content);
+                          handleSorting(item.content.key);
                         }
                       }}
                       dataTestIdPrefix={`${dataTestIdPrefix}_${
                         !isItemString(item)
-                          ? item?.content.toLowerCase().replace(/ /g, '_')
+                          ? item?.content.key.toLowerCase().replace(/ /g, '_')
```
kboutsika, line 239: Do we need  optional chaining operator here? item will always exists, right?

kboutsika (approved):

src/components/Table/Table.tsx
```diff
@@ -223,20 +223,20 @@ function Table<T>({
...
                       }
                       component={'th'}
-                      key={`${isItemString(item) ? item : item.content}`}
+                      key={`${isItemString(item) ? item : item.content.key}`}
                       sticky={fixedHeader}
                       padded={padded}
                       width={columnsWithWidth[index] ? `${columnsWithWidth[index]}%` : 'initial'}
                       isSortable={!isItemString(item) && item.isSortable}
-                      isActive={!isItemString(item) ? item.content === sorting.column : false}
+                      isActive={!isItemString(item) ? item.content.key === sorting.column : false}
```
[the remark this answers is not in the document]
whosyourtaco: Nice catch! I don't think the Boolean() cast is necessary though.

src/components/Table/Table.test.tsx
whosyourtaco: I was actually contemplating whether or not to use something more generic like `key` or make it more specific, like `sortingKey`. But since it is used only for the sorting functionality and some test ids, I guess it makes sense to name it `sortingKey`. Ty!

src/components/Table/Table.tsx
whosyourtaco: Nice catch! 😊

src/components/Table/Table.tsx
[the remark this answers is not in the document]
michalisKout: Yep, I added just in case you need a strict boolean result, otherwise, you should be fine. 👍🏻

PR #330, opened 2021-07-14
fix(build): small fix after merge for jsx

panvourtsis:
## Description

removing all jsx pragma that left after merge and now is causing problem on building the next release


whosyourtaco (approved):
🎉

PR #336, opened 2021-07-27
[DS-220] table header and expandable rows misalignment

kostasdano:
## Description

[DS-220](https://orfium.atlassian.net/browse/DS-220)

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -81,7 +81,10 @@ function Table<T>({
 
   const [sorting, setSorting] = useState<Sort>(initialSort);
 
-  const columnCount = onCheck ? columns.length + 1 : columns.length;
+  let columnCount = onCheck ? columns.length + 1 : columns.length;
+
+  const hasExpandableRows = data.some(row => !!row.expanded);
+  columnCount = hasExpandableRows ? columnCount + 1 : columnCount;
 
   useEffect(() => {
     if (onSort) {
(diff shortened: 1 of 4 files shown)


src/components/Table/Table.tsx
```diff
@@ -81,7 +81,10 @@ function Table<T>({
 
   const [sorting, setSorting] = useState<Sort>(initialSort);
 
-  const columnCount = onCheck ? columns.length + 1 : columns.length;
+  let columnCount = onCheck ? columns.length + 1 : columns.length;
+
+  const hasExpandableRows = data.some(row => !!row.expanded);
```
whosyourtaco, line 86: Probably better to use `Boolean(row.expanded)`
@panvourtsis 👀 👀

src/components/Table/Table.tsx
```diff
@@ -81,7 +81,10 @@ function Table<T>({
...
+  columnCount = hasExpandableRows ? columnCount + 1 : columnCount;
```
whosyourtaco, line 87: Better to do something like this instead of mutating
```
const getColumnCount = () => {
    if(!onCheck && !hasExpandableRows){
        return columns.length;
    } else if (onCheck && hasExpandableRows) {
        return columns.length + 2;
    } else {
        return columns.length + 1;
    }
}
```

whosyourtaco (approved):

src/components/Table/Table.tsx
panvourtsis: i agree and shall i suggest this as it defines what is boolean by name and type and has no dependency in other scopes

```
const getColumnCount = (columns, onCheck, hasExpandableRows) => {
    if(!Boolean(onCheck) && !hasExpandableRows){
        return columns.length;
    }
    if (Boolean(onCheck) && hasExpandableRows) {
        return columns.length + 2;
    }
    
    return columns.length + 1;
}
```

PR #337, opened 2021-07-30
[DS-229] Fix / Table header width when expandable

Christos-Zacharopoulos:
## Description

[ticket](https://orfium.atlassian.net/browse/DS-229)

## Screenshot

image

## Test Plan

Updated snaps

### DIFF
--- a/src/components/Table/Table.style.ts
+++ b/src/components/Table/Table.style.ts
@@ -6,7 +6,9 @@ import { Theme } from '../../theme';
 export const tableStyle = () => (): SerializedStyles =>
   css({ width: '100%', borderCollapse: 'collapse', tableLayout: 'fixed' });
 
-export const tableRowHeadersStyle = () => (theme: Theme): SerializedStyles =>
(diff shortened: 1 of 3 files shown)


kostasdano (approved):

src/components/Table/Table.style.ts
```diff
@@ -19,6 +21,6 @@ export const tableRowHeadersStyle = () => (theme: Theme): SerializedStyles =>
     },
 
     'th:last-child': {
-      paddingRight: theme.spacing.md,
+      paddingRight: hasExpandableRows ? 'inherit' : theme.spacing.md,
```
panvourtsis, line 24: here also work the `hasExpandableRows && theme.spacing.md

panvourtsis (approved):
LGTM 🎊

src/components/Table/Table.style.ts
Christos-Zacharopoulos: Tried it and TS throws an error.

`TS2769: No overload matches this call.`

PR #348, opened 2021-08-17
fix: [DS-237] change hover bg-color to filter

kboutsika:
## Description
[DS-237](https://orfium.atlassian.net/browse/DS-237)

Changed for `hover` state so as background color  to be 10% darken (or lighten) of the `default` color

## Screenshot
Στιγμιότυπο 2021-08-17, 4 54 40 μμ
Στιγμιότυπο 2021-08-17, 6 20 20 μμ

### DIFF
--- a/src/components/Filter/utils.ts
+++ b/src/components/Filter/utils.ts
@@ -1,6 +1,11 @@
+import { darken, lighten, transparentize } from 'polished';
+
+import { Theme } from '../../theme';
 import { colorShades, pickTextColorFromSwatches } from '../../theme/palette';
+import { ColorShapeFromComponent } from '../../utils/themeFunctions';
 import { defineBackgroundColor, stateBackgroundColor } from '../Button/utils';
 import { BackgroundColorProps, BaseColorProps, BorderProps, HoverBorderProps } from './types';
+
 
 export const FILTER_OPTIONS_MAX_HEIGHT = 253;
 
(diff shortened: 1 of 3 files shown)


panvourtsis:
design QA done ?

kboutsika:
> design QA done ?

no still need approval for design team

kboutsika:
@panvourtsis design QA done!

elektracodes (approved):

src/components/Filter/utils.ts
```diff
@@ -70,8 +75,18 @@ export const getHoverBorder = ({
...
   }
   if (styleType === 'outlined' || hasSelectedValue) {
-    return `${stateBackgroundColor(theme, 'hover', calculatedColor, true)}`;
+    return `${getHoverBackgroundColor(theme, calculatedColor)}`;
   }
 
   return 'transparent';
 };
+
+export const getHoverBackgroundColor = (
```
panvourtsis, line 84: consider use this function `stateBackgroundColor` as the logic is already there

kboutsika: thank you! 
`stateBackgroundColor` was initially used here but the hover background color was not the one we wanted, as inside `stateBackgroundColor` the background color for hover shade based on the primary or secondary color but here we need lighten or darken of the default color (that is primary color with shade 100)

Maybe I miss something tho, so if `stateBackgroundColor` can be used here, could you explain how? thanks again

kboutsika: @panvourtsis I have created an issue [here](https://github.com/Orfium/orfium-ictinus/issues/349), because i need this PR merged for the current sprint. Let' s discuss the improvement there.

Thank you!

panvourtsis (approved):

PR #350, opened 2021-08-23
feat: [DS-243] close modal on esc

kboutsika:
## Description

Added an option that defines if the modal should close with esc button or not, default value is true.

[DS-243](https://orfium.atlassian.net/browse/DS-243)


panvourtsis (approved):

panvourtsis:
LGTM 🎊 - very good job

PR #351, opened 2021-08-30
Feat/tooltip change

panvourtsis:
## Description

This PR tends to solve the tooltip issues from the previous library used such as miscalculation when absolute parent exists, many pre existing elements on DOM etc.

Things that i made:
- Changing from `react-tooltip` to `tippyjs`
- Fixing `tippyjs` for testing and mocking it
- Changing basic component with `forwardRef` for both future and for `tippyjs` to passing ref and works correctly

This is outside of the v4 roadmap but solves the bugs of our vital component.

Already tested and worked on:
- CMO-Portal-Client
- Video-Hunter
- CMO-usage-logs

Only issue found is on CMO-portal-client and specifically on `src/pages/Works/Works.style.ts` when it overrides the tooltip style to target a more popover functionality. @elektracodes @kostasdano 

## Screenshot

Nothing changed UI wised.

## Test Plan

Storyshots


SVonKarajann (approved):
LGTM

src/components/ButtonBase/ButtonBase.tsx
```diff
@@ -35,7 +35,11 @@ export type Props = {
   disabled?: boolean;
 };
 
-const ButtonBase: React.FC<Props & TestProps & EventProps & EventButtonProps> = props => {
+//@TODO fix props to not overwrite button props
```
elektracodes, line 38: Should we open a task for this?

PR #360, opened 2021-09-17
fix(button): change button to React.FC

elektracodes:
## Test Plan

This fixes the typescript errors on dist


elektracodes:
Wrong commit message

PR #375, opened 2021-09-29
Merging master branch to next

panvourtsis:
## Description

This PR is the merge of `master` to the `next` branch. 

I've tested all the components side by side with both master and next and i found out that there is a discrepancy on the filters component.

After discussing with the design team we came to a conclusion that we will revisit the filter with a new ticket for the V4. This ticket will include simpler logic and a more robust overall solution that will be the final for the upcoming filters.


Christos-Zacharopoulos (approved):
image

PR #387, opened 2021-11-04
Fix/color generation

panvourtsis:
## Description

This PR solves an issue that we had where when color was given to the color generator that was too dark or too light it was hitting a limit. That resulted to show only some of the shades e.g 50-650. This caused a problem when we wanted to access that color shade at 950 because it will not be there.

Now with the new calculation, there is a small change on some shades of some colors that I already mentioned to the design team and they will change

Additionally, I added the neutral black and white to the showcase now as we need to know them as system colors.

Lastly, I added the color generation with shades on color utility. This is a much-needed feature to know the calculations of any new color.

**changed colors**
```
dark grey
800
850

lightGrey
800

red
800

magenta
800
850

purple
800
850

blue
800

lightBlue
800
```

## Screenshot

utility
image

new neutral colors
image

## Test Plan

Added a snapshot test for all color palette.
Updated utils test for current colors and swatch because darkGrey 850 has changed.


whosyourtaco:
image

Is something wrong with the swatches calculation?

panvourtsis:
> image
> 
> Is something wrong with the swatches calculation? Maybe this comment is out of scope

if this is from the utility then swatches doesn't apply there

whosyourtaco (approved):

PR #388, opened 2021-11-04
fix: adds toLowerCase in string equality

whosyourtaco:
Remember all those times that the text returned black when it should have been white? All those times `primary` didn't work? Well, `#8c0d95 === #8C0D95` returns false :) *screams internally*


mkarajohn (approved):
10/10, would approve again

PR #389, opened 2021-11-08
fix: fix color calculations

whosyourtaco:
## Description

- Fix Filter color calculations by deprecating the `activeCalculatedColor` variable and only using the generated palette that is provided by the user through the `theme` prop (`calculatedColor`)

Before (secondary: `blue-600`): 
image

After (secondary: `blue-600`): 
image


panvourtsis (approved):
LGTM 🎊

PR #390, opened 2021-11-08
feat: select with final form debugging [WIP]

kostasdano:
## Description

This is a **WIP PR** to investigate the Select - ReactFinalForm integration bug


whosyourtaco:
Proposal of a different solution:

Revert Select to its original state, without any changes.

Add 
```        
onMouseDown={event => {
   event.preventDefault();
}}
``` on the ListItem component.

This will effectively prevent the onBlur method from running `onMouseDown` (it blurs by default) and will let `onClick` run (since in this case the onBlur event blocks the onClick event). The blur event will still happen, but only after `onClick` is finished.

kostasdano:
> Proposal of a different solution:
> 
> Revert Select to its original state, without any changes.
> 
> Add
> 
> ```
> onMouseDown={event => {
>    event.preventDefault();
> }}
> ```
> 
> on the ListItem component.
> 
> This will effectively prevent the onBlur method from running `onMouseDown` (it blurs by default) and will let `onClick` run (since in this case the onBlur event blocks the onClick event). The blur event will still happen, but only after `onClick` is finished.

nice approach! this actually works and it also prevents the awkward animation where the textfield label falls and then elevates back when an option is clicked

Christos-Zacharopoulos (approved):
After fiddling around with it, I couldn't find any reason not to go with @whosyourtaco proposal. 

Maybe add a github issue referencing this PR in order to revisit this in the future.

kostasdano:
Closing this PR in order to open a new one on a different branch and only with the ListItem.tsx change

PR #399, opened 2021-11-12
Feat/ds 260

whosyourtaco:
## Description

[DS-260]

- Implement Search Component based on TextInputBase component
- Add new stories for Search Field
- Designs based on [Figma](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System-V4---Working-File?node-id=10283%3A104365)
- Any suggestions for generifying existing css functions are welcome :) 

image

[DS-260]: https://orfium.atlassian.net/browse/DS-260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/SearchField/SearchField.stories.mdx
+++ b/src/components/SearchField/SearchField.stories.mdx
@@ -4,9 +4,9 @@ import SearchField from './SearchField'; import SearchFieldShowcase from './Sear
 
 <Meta title="Design System/SearchField" component={SearchField} />
 
-# TextField
+# SearchField
 
-A universal SearchField component that the input component for any forms.
+A Search field component based on the TextInputBase's styles.
 
 # Usage
 
--- a/src/components/SearchField/SearchField.stories.mdx
+++ b/src/components/SearchField/SearchField.stories.mdx
@@ -25,7 +25,7 @@ import { SearchField } from '@orfium/ictinus';
 Search Field and available sizes
 
 <Preview>
-  <Story name="Search Field with sizes">
+  <Story name="Search Field sizes">
     <SearchFieldShowcase />
   </Story>
 </Preview>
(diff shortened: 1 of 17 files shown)


panvourtsis:
good job @whosyourtaco. I would like to see this passing the design QA first

Some comments before that. 
The first one is that the state change isn't working for focus although it does when I use my mouse

image

The second one is that "Text Field Sizes" story needs to change as the search has two sizes. Make it more visible. Playgrounds have to have knobs.

Also please add the frame of figma as a link in the story (check others).

src/components/SearchField/SearchField.stories.mdx
```diff
@@ -0,0 +1,51 @@
+import { Meta, Preview, Props, Story } from '@storybook/addon-docs';
+import Stack from '../storyUtils/Stack';
+import SearchField from './SearchField'; import SearchFieldShowcase from './SearchFieldShowcase';
+
+<Meta title="Design System/SearchField" component={SearchField} />
+
+# TextField
```
panvourtsis, line 7: search field *

src/components/SearchField/SearchField.stories.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+A universal SearchField component that the input component for any forms.
```
panvourtsis, line 9: this doesnt make sense when i read it

src/components/TextField/commons.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { FC, MouseEvent } from 'react';
```
panvourtsis, line 1: can we move this to a components folder within the textfield. If we wnt to add more this will gets bigger and is not aligned with our components/file structure

https://orfium.atlassian.net/wiki/spaces/FE/pages/2147057733/Components+File+Structure

src/components/SearchField/SearchField.stories.mdx
whosyourtaco: Damn copy paste :P

Christos-Zacharopoulos: image

panvourtsis (approved):

PR #405, opened 2021-11-24
fix: correct filter border radius and paddings

whosyourtaco:
## Description

- Removes padding outside of Filter component
- Fixes border radius issues, as shown on the ticket [DS-284]

## Screenshots 

After the changes:
image
image

[DS-284]: https://orfium.atlassian.net/browse/DS-284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/List/NormalList/NormalList.style.ts
+++ b/src/components/List/NormalList/NormalList.style.ts
@@ -1,16 +1,20 @@
 import { css, SerializedStyles } from '@emotion/react';
+import { Theme } from 'theme';
 import { rem } from 'theme/utils';
 
 export const listStyle = ({
   width,
   height,
+  isSearchable,
 }: {
   width?: number;
   height?: number;
-}) => (): SerializedStyles => css`
+  isSearchable?: boolean;
+}) => (theme: Theme): SerializedStyles => css`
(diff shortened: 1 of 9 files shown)


src/components/List/NormalList/NormalList.style.ts
```diff
@@ -1,16 +1,20 @@
...
 }: {
   width?: number;
   height?: number;
-}) => (): SerializedStyles => css`
+  isSearchable?: boolean;
+}) => (theme: Theme): SerializedStyles => css`
   padding-left: 0;
   margin-top: 0;
   margin-bottom: 0;
+  border-radius: ${isSearchable ? 'initial' : theme.spacing.xsm};
```
panvourtsis, line 17: why we have special cases for radius ?

whosyourtaco: @panvourtsis Because it causes this to happen 
image

panvourtsis (approved):

PR #408, opened 2021-11-30
fix: some of the security warnings

panvourtsis:
## Description

This PR solves some security warnings by upgrading storybook and svgr. In order to work well with the latest svgr webpack 5 was applied

* nth-check
* js-yaml


Christos-Zacharopoulos (dismissed):

Christos-Zacharopoulos:
These tests are still failing

PR #409, opened 2021-12-01
Perf/table patch render

panvourtsis:
## Description

This PR focus on fixing some performance issues of the Table component. 

1) cell rerender even if it has the same content
2) table rerender with the same data set

### The problem
The data that I was filtering had space in it. So with one space pressed nothing was changed on our dataset yet the table component and the table cells were rendered again.

Screenshot 2021-12-01 at 9 57 48 AM
Screenshot 2021-12-01 at 9 57 27 AM
perf-problem

### The solution
The solution has two parts. Firstly I removed the `columnsHasNumberArr` from the `TableCell` component as we can do that calculation inside per content. This resulted in the following.
Screenshot 2021-12-01 at 10 01 09 AM

Then I introduced the `react-fast-compare` [library](https://www.npmjs.com/package/react-fast-compare) a faster alternative to `fast-deep-equal` for deep equal Table props. This way we can ensure that our component not have impact to any application that will be used. This doesn't mean that the user will be free to use it with no memorization on `columns` or `data` passed but will secure a big part of it.
perf-solution
Screenshot 2021-12-01 at 9 58 59 AM

## Test Plan

React profiler
New case on mdx


Christos-Zacharopoulos (approved):
image

PR #411, opened 2021-12-02
Hotfix/truncation tooltip

whosyourtaco:
## Description

- Extend Table types to include `tooltipContent` as a required prop, when content is JSX
- Added `offsetWidth` to dependency array, so that the tooltip calculation updates on zoom/res change
- Removed `shouldAlwaysShow` prop as it was unnecessary


panvourtsis:
Why this is in a draft mode? If its idle we can close it

Christos-Zacharopoulos:
@whosyourtaco Is this going to move forward or should we close it?

panvourtsis:
@whosyourtaco I am closing this since there was no update. If something change we will reconsider it

PR #415, opened 2021-12-06
[DS-224]: style(elevation): elevation updated

SVonKarajann:
## Description

[Ticket](https://orfium.atlassian.net/browse/DS-224)


panvourtsis (approved):
LGTM 🎊  and good catch on blur

PR #422, opened 2021-12-10
[DS-82] Fix/avatar shapes

Christos-Zacharopoulos:
## Description

ticket: [DS-82](https://orfium.atlassian.net/browse/DS-82)

## Screenshot
before
image

after
image

## Test Plan

updated snaps


src/components/Avatar/Avatar.tsx
```diff
@@ -31,7 +31,7 @@ export type Props = {
 };
 
 export type AvatarSizes = 'xxxs' | 'xxs' | 'xs' | 'sm' | 'md' | 'lg';
-export type AvatarShapes = 'regular' | 'rounded' | 'circular';
+export type AvatarShapes = 'circular';
```
SVonKarajann, line 34: is AvatarShapes type, and shape prop needed?

SVonKarajann (approved):

src/components/Avatar/Avatar.tsx
Christos-Zacharopoulos: As discussed off line I will not remove this in order to avoid a breaking change. They will be removed in v5 of Ictinus. 

Marking this as resolved.

PR #423, opened 2021-12-13
[DS-281]: feat: getAATextColor function added

SVonKarajann:
## Description

[Ticket](https://orfium.atlassian.net/browse/DS-281)


panvourtsis:
Can we have the pale and color util fixed as well with this solution? 
Screenshot 2021-12-14 at 3 22 29 PM

src/theme/palette.ts
```diff
@@ -117,76 +119,22 @@ export type formFieldStyles = 'filled' | 'outlined' | 'elevated';
...
-    500: ['darkGrey', 'magenta', 'purple', 'darkBlue', 'blue', 'darkOrange'],
-    450: ['darkGrey', 'purple', 'darkBlue', 'blue', 'darkOrange'],
-    400: ['darkGrey', 'darkOrange'],
-    350: ['darkGrey', 'darkOrange'],
-  };
-  const pickedShade = colorsForWhiteText[shade];
-  const pickedColor =
-    (pickedShade && pickedShade?.find(item => item === color)) || color === 'neutralBlack';
+): string => {
+  const palette = enhancePaletteWithShades(lightPaletteConfig);
```
Christos-Zacharopoulos, line 128: I am a little concerned about performance here. We are calling `enhancePaletteWithShades` that's a recursive function two times in a row.

Christos-Zacharopoulos: Took a quick look at the usages and where we use `pickTextColorFromSwatches` we have access to the theme. 

I suggest we pass the theme as a prop to `pickTextColorFromSwatches` and avoid the overhead of running `enhancePaletteWithShades` over and over again.

Christos-Zacharopoulos (changes requested):

panvourtsis:
@Christos-Zacharopoulos can I suggest trying anything new on a different scope? I mean that we are trying to remove the mapping of the colors and replace it with AA. We can create a new ticket with a scope to make this function more performant.  What do you say?

src/theme/palette.ts
panvourtsis: @Christos-Zacharopoulos can I suggest trying anything new on a different scope? I mean that we are trying to remove the mapping of the colors and replace it with AA. We can create a new ticket with a scope to make this function more performant. What do you say?

Christos-Zacharopoulos: Works for me. 👍

Christos-Zacharopoulos (approved):
LGTM! 🎉

PR #425, opened 2021-12-14
[DS-303]: Feat/filter base split

whosyourtaco:
## Description

- Extracts FilterBase from Filter to separate component file
- Adds functionality for opening a DatePicker through a Filter
- UI Changes to align components with Figma
- Updated `calendar-filled` icon. It was originally exported in a wrong way


src/components/DatePicker/DatePickInput/DatePickInput.tsx
```diff
@@ -13,12 +16,16 @@ const ON_CHANGE_MOCK = () => {};
...
-  handleClear: (event: React.KeyboardEvent) => void;
+  handleClear: (event?: React.KeyboardEvent) => void;
+  filterType?: FilterType;
   isRangePicker: boolean;
   selectedDay: Range;
   /** Props for styling the input */
   inputProps?: TextFieldProps;
   dateFormatOverride?: DateFormatType;
+  open: boolean;
+  buttonType?: 'primary' | 'secondary';
```
panvourtsis, line 27: i am very skeptical regarding these props `buttonType, styleType, filterType` as these are referring only to filter. I would suggest having them as one object that have them all. 

Any ideas here @elektracodes @Christos-Zacharopoulos ?

src/components/Filter/components/FilterBase/FilterBase.tsx
```diff
@@ -0,0 +1,136 @@
...
+  valueSpanStyle,
+  wrapperStyle,
+} from '../../Filter.style';
+import { FilterType, StyleType } from '../../types';
+
+// eslint-disable-next-line react/display-name
+export const FilterBase = forwardRef<
+  HTMLButtonElement,
+  {
+    dataTestId?: string;
```
panvourtsis, line 25: i would expect these to be a type like we do it

src/components/Filter/components/FilterBase/FilterBase.tsx
```diff
@@ -0,0 +1,136 @@
...
+  childrenWrapperStyle,
+  dividedButtonStyle,
+  divider,
+  labelSpanStyle,
+  valueSpanStyle,
+  wrapperStyle,
+} from '../../Filter.style';
+import { FilterType, StyleType } from '../../types';
+
+// eslint-disable-next-line react/display-name
```
panvourtsis, line 21: please define the display name for future debugging

src/components/Filter/components/FilterBase/FilterBase.tsx
panvourtsis: i would also expect these to be the same for filter. Because there you use these props as well

src/components/Filter/components/FilterBase/FilterBase.tsx
```diff
@@ -0,0 +1,136 @@
...
+    disabled: boolean;
+    handleOpen: () => void;
+    onClear: () => void;
+    selectedItemLabel?: string;
+    open: boolean;
+    hasSelectedValue: boolean;
+    label?: string;
+    iconName: AcceptedIconNames;
+    iconColor: string;
+    iconSize?: number;
```
panvourtsis, line 36: why we need this ? dont we have same sizes for icons ? if no why (can we ask the design team)?

src/components/Filter/components/FilterBase/FilterBase.tsx
```diff
@@ -0,0 +1,136 @@
...
+    calculatedColor,
+    buttonType,
+    disabled,
+    open,
+    styleType,
+    hasSelectedValue,
+    filterType,
+  };
+
+  const pickIconColor = () => {
```
panvourtsis, line 77: now that we move this can we use a `useCallback` here?

src/components/DatePicker/DatePickInput/DatePickInput.tsx
```diff
@@ -27,50 +34,97 @@ const DatePickInput = React.forwardRef<HTMLInputElement, Props & InputProps>(
...
-        {isRangePicker ? (
-          <TextField
-            ref={ref}
-            {...inputProps}
-            onFocus={handleFocus}
-            onKeyDown={handleClear}
-            onChange={ON_CHANGE_MOCK}
-            placeholder="Date (start) - Date (end)"
-            value={
+    const renderBase = React.useMemo(() => {
```
panvourtsis, line 56: using a useMemo here with all the properties taken from the parent component will get us nothing in return. We can skip that computation

src/components/DatePicker/DatePickInput/DatePickInput.tsx
panvourtsis: another approach could be to have a context on the `FilterBase` to pass these values and then `Filter`, `Datepicker` and `DatePickInput` will read these values if under that context from provider. Let me know your thoughts @Orfium/front-end

panvourtsis (changes requested):

src/components/DatePicker/DatePickInput/DatePickInput.tsx
whosyourtaco: I feel like adding a context would be too much to accomplish what we want. The object idea sounds good to me!

src/components/DatePicker/DatePicker.tsx
```diff
@@ -36,6 +37,12 @@ export type Props = {
   isClearable?: boolean;
   /** if the datepicker's default date is today instead of placeholder text */
   isDefaultNow?: boolean;
+  /** This property defines whether the DatePicker component's parent is a Select or a Filter and the Filter's type */
+  filterType?: FilterType;
+  /** The type of the filter button's palette - defaults to "primary" */
+  buttonType?: 'primary' | 'secondary';
+  /** Defines the style type of the filter button */
+  styleType?: StyleType;
```
kostasdano, line 45: maybe you could add the filterConfig object prop here too

kostasdano (approved):
🎉

kostasdano (approved):

src/components/Filter/Filter.style.ts
```diff
@@ -183,26 +183,23 @@ export const buttonStyle = (props: ButtonStyleProps) => (theme: Theme) => {
   };
 };
 
-export const childrenWrapperStyle = () => (theme: Theme) => {
+export const childrenWrapperStyle = () => () => {
   return {
+    lineHeight: '15px',
```
Christos-Zacharopoulos, line 188: 👀

src/components/DatePicker/DatePickInput/DatePickInput.tsx
Christos-Zacharopoulos: Sorry for being late. I was on PTO.

I prefer the object as prop. If we go to a context solution I would prefer to consider it after global states are implemented, as ther context described above is a kind of global behavior thingy thing.

Something like `filterBaseProps` like material does with it's components.

PR #431, opened 2021-12-16
[DS-298]: feat(select): add triangle icons

SVonKarajann:
## Description

[Ticket](https://orfium.atlassian.net/browse/DS-298)

Also 
- we resolve [this bug](https://orfium.atlassian.net/browse/DS-298?focusedCommentId=24169)
- we make sure all listItems have ```cursor: pointer;``` property
- we fixed the disabled option bug
select-cursor-disabled

## Screenshot
select-icons

### DIFF
--- a/src/components/Select/Select.style.ts
+++ b/src/components/Select/Select.style.ts
@@ -34,7 +34,6 @@ export const rightIconContainer = (
   isSearchable: boolean
 ) => (): SerializedStyles => css`
   display: flex;
-  gap: ${rem(25)};
   cursor: pointer;
   transform: rotate(${open && !isSearchable ? '180' : '0'}deg);
   ${transition(0.2)}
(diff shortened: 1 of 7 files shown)


src/components/Select/Select.style.ts
```diff
@@ -27,3 +28,14 @@ export const selectWrapper = ({
     cursor: ${!isSearchable && 'pointer'};
   }
 `;
+
+export const rightIconContainer = (
+  open: boolean,
+  isSearchable: boolean
+) => (): SerializedStyles => css`
+  display: flex;
+  gap: ${rem(25)};
```
Christos-Zacharopoulos, line 37: Can we discuss this with Design team to maybe use here a value from spacing?

SVonKarajann: not sure gap (grid-gap) property has any effect on components with ```display: flex;```. Just gonna remove that

Christos-Zacharopoulos (approved):

panvourtsis (approved):

PR #433, opened 2021-12-20
fix: Set button type to 'button' by default

### LINKED ISSUE #419: Set the default button type to 'button'
Considering the vast majority of the usecases of the buttons inside orfium React applications, it would make sense for the default button type to be set to `button` instead of leaving it blank and having it default to `submit`

mkarajohn:
## Description

Resolves #419 . Snapshots had to be updated too.


SVonKarajann (approved):

Christos-Zacharopoulos:
image

SVonKarajann (approved):

PR #438, opened 2022-01-04
[DS-287]: Feat/avatar update

SVonKarajann:
## Description

[Ticket](https://orfium.atlassian.net/browse/DS-287)

## Screenshot
avatars
avatar-stack


src/components/Avatar/AvatarStack/AvatarStack.style.ts
```diff
@@ -0,0 +1,27 @@
+import { css, SerializedStyles } from '@emotion/react';
+import { rem } from 'theme/utils';
+
+import { AvatarSizes } from '../Avatar';
+
+const iconMarginLeft = (size: AvatarSizes) => {
+  switch (size) {
+    case 'lg':
+      return rem(-8);
```
Christos-Zacharopoulos, line 9: 🤔

Christos-Zacharopoulos: Why are negative values used here? Can we find a cleaner workaround here?

src/components/Avatar/AvatarStack/AvatarStack.tsx
```diff
@@ -0,0 +1,65 @@
...
+    {
+      maxAvatars = 4,
+      size = 'md',
+      color = 'lightGrey-600',
+      dataTestId = '',
+      children: childrenProp,
+    },
+    ref
+  ) => {
+    const clampedMax = maxAvatars < 1 ? 1 : maxAvatars;
```
Christos-Zacharopoulos, line 33: It's better to throw Error if `maxAvatars` is 0 or a negative.

src/components/Avatar/Avatar.style.ts
```diff
@@ -51,10 +45,13 @@ export const avatarStyle = ({
   width: ${sizeBasedOnProp(size)};
   height: ${sizeBasedOnProp(size)};
   border-radius: 100%;
+  border: ${rem(1)} solid ${theme.utils.getColor('lightGrey', 100)};
+  box-sizing: border-box;
```
mkarajohn, line 49: Not relevant to this, but I think we are not consistent in applying this to all of our components. Shouldn't we be?

src/components/Avatar/AvatarStack/AvatarStack.style.ts
mkarajohn: @Christos-Zacharopoulos They are overlapping here, so negative margins are one way to do this. Would you suggest doing it another way?

src/components/Avatar/AvatarStack/AvatarStack.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    const extraAvatars = children.length > clampedMax ? children.length - clampedMax : 0;
+
+    return (
+      <div
+        ref={ref}
+        data-testid={generateTestDataId('avatarstack', dataTestId)}
+        style={{ display: 'flex' }}
+      >
+        {children.slice(0, children.length - extraAvatars).map((child, index) => {
```
mkarajohn, line 45: In general we can be memoizing return values of array operations, like `slice` or `map`, instead of always iterating over the array, even when we know the returned values will be the same. 

Just something to keep in mind, nonblocking.

mkarajohn (approved):

src/components/Avatar/AvatarStack/AvatarStack.style.ts
Christos-Zacharopoulos: The overlap is handled by the z-index. The problem is the positioning of the items. Right now they are placed outside of their parent container. Also,

1) the first one is outside the container due to the negative margin. 
image
This is easily fixed.

2) This has the disadvantage of having the margin of each size of them hard-copied. This is easily forgotten and might create issues in the future. 

3) Using negative values is like trying to find a dirty workaround for a problem. Each element can use it's dedicated space without having to do this. I prefer to use this kind of solutions for styling third party libraries that I absolutely need and can't find an alternative.

Proposed way:

This might need a few more changes but it's cleaner and helps safeguard future changes.

We can define the width instead of the margin-left in `avatarStackStyle` of all the containers except for the last one. This way everything is nice and tidy inside their container. 

I would also like to see more consistency in the amount of overlap each items has.

For example: 
`lg` size is 46 but overlap is 8.
`md` size is 24 but overlap is 4.
`sm` size is 20 but overlap is 4.
`xs` size is 16 but overlap is 2.

There is no rule or consistency in the amount of overlap. We could propose to the design team for the overlap to be 1/3 or 1/4 of the avatar's size. And it's also very easy.

In `Avatar.styles.ts`  there is a `sizeBasedOnProp`  function. 

If you refactor this to 
`const sizeBasedOnProp = (size: AvatarSizes) => {
  switch (size) {
    case 'md':
      return 24;
    case 'sm':
      return 20;
    case 'xs':
      return 16;
    default:
      return 46;
  }
};`

then you can use this to also set the width of the `avatarStackStyle`

src/components/Avatar/AvatarStack/AvatarStack.style.ts
```diff
@@ -1,20 +1,22 @@
...
-  switch (size) {
-    case 'lg':
-      return rem(-8);
-    case 'xs':
-      return rem(-2);
-    default:
-      return rem(-4);
-  }
-};
+const OVERLAP_FACTOR = 0.8;
```
Christos-Zacharopoulos, line 8: has this been discussed with the design team?

src/components/Avatar/AvatarStack/AvatarStack.tsx
```diff
@@ -30,27 +30,35 @@ const AvatarStack = React.forwardRef<HTMLDivElement, Props & TestProps & DivProp
...
-          return (
-            <div key={index} css={avatarStackStyle({ zIndex: children.length - index, size })}>
-              {child}
-            </div>
-          );
-        })}
+        {renderContent()}
         {extraAvatars ? (
-          <div css={avatarStackStyle({ zIndex: 0, size })}>
+          <div css={avatarWrapperStyle({ zIndex: 0, size })}>
```
Christos-Zacharopoulos, line 61: zindex is no longer needed

SVonKarajann: why not?

Christos-Zacharopoulos: image
My mistake. I thought the overlap was the other way as shown in the picture above. Double checked it.

src/components/Avatar/AvatarStack/AvatarStack.style.ts
SVonKarajann: yeap

Christos-Zacharopoulos (approved):
image

PR #441, opened 2022-01-13
[DS-305] Feat/global states

Christos-Zacharopoulos:
## Description

ticket: [DS-305](https://orfium.atlassian.net/browse/DS-305)

This PR started as draft to discuss the implementation direction taken for the global states styling. 

First the button was implemented and checked, and then a series of PRs were made to allow for easier code reviews for each component. 

Bellow you can see the initial analysis on global states and implementation decisions on Button and you can check each PR for every component.

**_Initial Analysis below_** 

This is a first though of the global states implemented and applied to button.

States are:
<html>
<body>

style/state | Hover | Focused | Pressed | Visited | Visited & Focused | Disabled
-- | -- | -- | -- | -- | -- | --
Background | {background-color: ( color: default lightGrey, shade: default 0 ) => getColor(color, shade +50)(mind the extra dark shades)} | default or custom | {background-color: ( color: default lightGrey, shade: default 0 ) => getColor(color, shade +100)(mind the extra dark shades)} | no ruleonly filters | no ruleonly filters | Default with 50% opacity
Border | default or  custom per component | +1 px color Magenta 500{border: (width: CONST) => width +1 solid getColor( magenta, 500)} | default or  custom per component | no ruleonly filters | no ruleonly filters | Default with 50% opacity
Color | swatch or custom | swatch or custom | swatch or custom | no ruleonly filters | no ruleonly filters | Default with 50% opacity
Icon | swatch or custom | swatch or custom | swatch or custom | no ruleonly filters | no ruleonly filters | Default with 50% opacity

</body>
</html>

#### Config: 

In order to make easy changes and for readability I extracted all the values of the global rules to a configuration file.

#### Hover:
Hover changes only the background:

- to `lightGrey-50` by default or
- to whatever color we have passed+50.  If we go out of bounds ( > 950 ) then we go one step down.

At the beginning I was returning only a string with the new color  but I decided to return an object to allow for future expansion of the hover state global rule.

#### Focus:
Hover changes only the border to `Magenta-500`. 

By adding the color to the border on components that don't have one we had the issue of the component expanding in width and looking kinda buggy like below.

buggy

So I used outline:
noice

The response here is a little bigger. I return the `styleOutline` and `styleBorder` ( in case we need/want it) as well as the values alone ( `borderWidth` and `focusColor` ) in case we can't wrap a component in a single container.

Also, in button I used the `:focus-visible` pseudo class to avoid having the items being focused when we click them. I thing it's got a fairly good browser support ( https://caniuse.com/?search=focus-visible ) and we can be confident in using it.

#### Pressed:

Pressed changes only the background:

- to `lightGrey-100` by default or
- to whatever color we have passed+100.  If we go out of bounds ( > 950 ) then we go two steps down.

At the beginning I was returning only a string with the new color but I decided to return an object to allow for future expansion of the pressed state global rule.

Since this is pretty similar to the hover rule, I abstracted the shade functionality to a separate function in order to reuse it.

#### Disabled:

The disabled rule is global and pretty easy. We use the default state with an opacity of 50% and I also added the cursor.  I added the style ready to use and again , if for any reason we don't want to use the style as is , each property is returned individually. 

### Changes that also happened along the way:

While integrating the global states to the button component I tried to also make the code a little easier to maintain and read.

So:

- I added a button configuration file and moved some constant values there.
- I used the `:disabled` and `:not(:disabled)` pseudo classes in order to not have ternary expressions for styling and keep it simple CSS.
- Made minor modifications to existing functions and removed redundant code.
- Added constants for values that re reused or that where hard to read (like `isOutlined`)

## Test Plan

Test all state functions  and their utils for all possible responses.


whosyourtaco:
@Christos-Zacharopoulos awesome job! this will solve many problems and prove to be a very useful addition in the long run! 🎉

src/components/Button/config.ts
```diff
@@ -0,0 +1,49 @@
+import { colorShades, flatColors } from '../../theme/palette';
+
+type ButtonConfig = {
+  sizes: {
+    [key in 'lg' | 'md' | 'sm' | 'default']: number;
```
panvourtsis, line 5: this and fontSize could be done like flatColors typing using `as const`

src/components/ButtonBase/ButtonBase.style.ts
```diff
@@ -1,47 +1,29 @@
...
-      return rem(56);
-    case 'sm':
-      return rem(36);
-    default:
-      return rem(46);
-  }
-};
+/** Calculates the button specific height based on the size passed to it **/
+export const heightBasedOnSize = (size: 'lg' | 'md' | 'sm') =>
+  buttonConfig.sizes[size] || buttonConfig.sizes.default;
```
panvourtsis, line 14: we lose the `rem` now ?

src/theme/states/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+  step: number;
+};
+
+export const getShadeWithStep = ({
+  shade = 0,
+  step = 50,
+}: GetShadeWithStepProps): typeof colorShades[number] => {
+  let calculatedShade = shade;
+
+  if (shade + step > 950) {
```
panvourtsis, line 14: this 950 must be documented in another variable and use from there. This is used in filters as well and might be useful as we go

panvourtsis:
Nice work @Christos-Zacharopoulos 🚀 , this makes a lot of sense. I would love to see a draft of this on TextFields as well maybe on a separate branch made from this?

src/components/Button/config.ts
Christos-Zacharopoulos: Also moved it to button base. Don't know why I created it on Button

docs/guides/GLOBAL_STATES.stories.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+## Pressed
+
+When pressed, the background color is darkened by two steps in shade ( +100 ).
+
+
+- If no background color exists, transparent or white, then on pressed we use lightGrey-100
+- If it goes out of bounds, it's lightened by two steps ( -100 )
+
+## Pressed
```
SVonKarajann, line 38: you ve added Pressed twice

docs/system/UTILS.stories.mdx
```diff
@@ -90,3 +91,191 @@ const calculatedColor = calculateColorBetweenColorAndType(color, type);
...
+...
+```
+
+---
+
+
+### getAAColor
+
+This util is provided from core `theme` or you can use it as a standalone function.
+In case of standalone you must provide the [Palette](/?path=/story/system-theme--page#theme-palette).
```
SVonKarajann, line 139: link doesnt work, it should be `[Palette](/story/system-theme--page#theme-palette)`

docs/system/UTILS.stories.mdx
```diff
@@ -90,3 +91,191 @@ const calculatedColor = calculateColorBetweenColorAndType(color, type);
 
 console.log(calculatedColor.color, calculatedColor.shade);
 ```
+
+---
+
+### getAAColorFromSwatches
+
+This util is provided from core `theme` or you can use it as a standalone function.
+In case of standalone you must provide the [Palette](/?path=/story/system-theme--page#theme-palette).
```
SVonKarajann, line 100: link doesnt work, it should be `[Palette](/story/system-theme--page#theme-palette)`

docs/system/UTILS.stories.mdx
```diff
@@ -90,3 +91,191 @@ const calculatedColor = calculateColorBetweenColorAndType(color, type);
...
+{
+  opacity: number;
+  cursor: string;
+}
+```
+
+#### Usage <br/><br/>
+
+```
+cursor: getDisabled({theme, color: 'blue', shade: 500}).cursor;
```
SVonKarajann, line 277: you say above that it gets 0 params.

Christos-Zacharopoulos: Guilty of copy-pasting the above as a template. Will fix.

src/components/Chip/Chip.style.ts
```diff
@@ -11,38 +12,62 @@ export const chipStyle = ({
...
+    ${flexCenterVertical};
+    height: ${theme.spacing.lg};
+    border-radius: ${theme.spacing.lg};
+    font-size: ${theme.typography.fontSizes['12']};
+    font-weight: ${theme.typography.weights.medium};
+    line-height: normal;
+    box-sizing: border-box;
+    padding: ${theme.spacing.xsm} ${theme.spacing.sm};
+    background-color: ${customFilled ? theme.utils.getColor(fill, 50) : theme.palette.white};
+    color: ${theme.utils.getColor('darkGrey', 850)};
```
SVonKarajann, line 29: Maybe use getAAColorFromSwatches based on background-collor here

SVonKarajann (approved):

PR #442, opened 2022-01-14
[DS-305] Feat / Global States - integration with text field

Christos-Zacharopoulos:
## Description

Integration of global states to text input base. ( duh, it says in tittle dum dum ) 

This uses the same styles and patterns described in https://github.com/Orfium/orfium-ictinus/pull/441

Major changes that happened in this PR:

- We only support outlined variant. I have not yet removed the types as this is a breaking change, but all types return the outlined version. 
- Abstracted part of the styling for the text input base component in a config. 
- Refactored the way we style text input base as it was a bit complicated.
- Focused state is the same as pressed/active state. When you focus an input you can write.

In general, no further changes where needed in the global states functions. I did the bare minimum in order to not get outside the scope of the ticket. I could have also done other tickets regarding text input but it would make it harder to review and distinguish what is changed for what ticket.

## Screenshot

Here you can see the new styles.. From left to right, default, hovered and pressed-focused.
image

## Test Plan

Only snaps where updated.

### DIFF
--- a/src/components/TextInputBase/TextInputBase.style.ts
+++ b/src/components/TextInputBase/TextInputBase.style.ts
@@ -13,7 +13,6 @@ export const flexContainer = () => (): SerializedStyles =>
     display: 'flex',
     alignItems: 'center',
     position: 'relative',
-    width: '-webkit-fill-available',
   });
 
 const borderConfig = textInputConfig.types.outlined.border;
--- a/src/components/TextInputBase/TextInputBase.style.ts
+++ b/src/components/TextInputBase/TextInputBase.style.ts
@@ -3,40 +3,59 @@ import { Theme } from 'theme';
 import { rem } from 'theme/utils';
 import { DEFAULT_SIZE, getTextFieldSize } from 'utils/size-utils';
 
+import { getDisabled, getHover, getPressed } from '../../theme/states';
+import { textInputConfig } from './config';
 import { Props } from './TextInputBase';
 import { LABEL_TRANSFORM_LEFT_SPACING } from 'components/Label/Label.style';
 
+export const flexContainer = () => (): SerializedStyles =>
+  css({
+    display: 'flex',
+    alignItems: 'center',
+    position: 'relative',
+    width: '-webkit-fill-available',
+  });
+
+const borderConfig = textInputConfig.types.outlined.border;
+
 const wrapperStyleSwitch = (
   theme: Theme,
+  dark?: boolean,
   lean?: boolean,
   error?: boolean,
-  styleType?: string,
   disabled?: boolean
 ) => {
   if (lean) {
-    return `
-    `;
+    return {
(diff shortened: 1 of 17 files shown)


Christos-Zacharopoulos:
image

src/components/TextInputBase/TextInputBase.tsx
```diff
@@ -3,9 +3,10 @@ import React, { FC } from 'react';
 import { formFieldStyles } from 'theme/palette';
 import { DEFAULT_SIZE } from 'utils/size-utils';
 
-import { errorMsgStyle, textFieldStyle, wrapperStyle } from './TextInputBase.style';
+import { errorMsgStyle, flexContainer, textFieldStyle, wrapperStyle } from './TextInputBase.style';
 import Icon from 'components/Icon';
 import { AcceptedIconNames } from 'components/Icon/types';
+import { textInputSizes } from './config';
```
SVonKarajann, line 9: getting an eslint(import/order) error regarding this line

src/components/TextInputBase/TextInputBase.style.ts
```diff
@@ -3,40 +3,59 @@ import { Theme } from 'theme';
...
+import { textInputConfig } from './config';
 import { Props } from './TextInputBase';
 import { LABEL_TRANSFORM_LEFT_SPACING } from 'components/Label/Label.style';
 
+export const flexContainer = () => (): SerializedStyles =>
+  css({
+    display: 'flex',
+    alignItems: 'center',
+    position: 'relative',
+    width: '-webkit-fill-available',
```
mkarajohn, line 16: Not a fan of using a non standard value for width here. Can we make this work in a standards compliant way? See here for standardized values: https://developer.mozilla.org/en-US/docs/Web/CSS/width#browser_compatibility

src/components/TextInputBase/TextInputBase.style.ts
```diff
@@ -48,75 +67,31 @@ export const wrapperStyle = ({
...
+        : theme.utils.getColor(borderConfig.color.default.name, borderConfig.color.default.shade)
+    }`,
+    borderRadius: isSearch ? rem(100) : theme.spacing.xsm,
+    flex: '1 1 100%',
+    userSelect: 'none',
+    position: 'relative',
+    opacity: disabled ? getDisabled().opacity : 1,
+    cursor: disabled || locked ? getDisabled().cursor : 'auto',
+    /** This is added to prevent the field from growing/shrinking when the Clear icon shows/hides.
+        The values used are the minimum widths of this field */
```
mkarajohn, line 91: Where can I see an example of this behaviour?

Christos-Zacharopoulos: This part was not refactored as it's completely out of the scope of the ticket.

mkarajohn (approved):

PR #443, opened 2022-01-17
[DS-302]: Feat/overlay

SVonKarajann:
## Description

[Ticket](https://orfium.atlassian.net/browse/DS-302)

### DIFF
--- a/src/components/Overlay/Overlay.stories.mdx
+++ b/src/components/Overlay/Overlay.stories.mdx
@@ -2,6 +2,7 @@ import { Meta, Story, Preview, Props } from '@storybook/addon-docs';
 import { withKnobs, boolean, array, select, text } from '@storybook/addon-knobs';
 import Overlay from './Overlay';
 import OverlayShowcase from '../storyUtils/OverlayShowcase';
+import Stack from '../storyUtils/Stack';
 
 <Meta title="Design System/Overlay" component={Overlay} />
 
@@ -24,7 +25,20 @@ import { Overlay } from '@orfium/ictinus';
 # Overlay
 
 <Preview>
-  <Story name="Overlay" parameters={{ decorators: [withKnobs] }}>
-    <OverlayShowcase anchor={select( 'anchor', ['left', 'top', 'right', 'bottom'],'left')} size={text('size', 33)}/>
+  <Story name="Overlay">
+    <Stack>
+      <OverlayShowcase anchor={'left'} size={33}/>
+      <OverlayShowcase anchor={'top'} size={33}/>
+      <OverlayShowcase anchor={'right'} size={33}/>
+      <OverlayShowcase anchor={'bottom'} size={33}/>
+    </Stack>
+  </Story>
+</Preview>
+
+# Overlay Playground
+
+<Preview>
+  <Story name="Overlay Playground" parameters={{ decorators: [withKnobs] }}>
+    <OverlayShowcase anchor={select( 'anchor', ['left', 'top', 'right', 'bottom'],'left')} size={text('size', 33)} playground/>
   </Story>
 </Preview>
--- a/src/components/Overlay/Overlay.stories.mdx
+++ b/src/components/Overlay/Overlay.stories.mdx
@@ -15,7 +15,7 @@ A universal Overlay component.
 ```js
 import { Overlay } from '@orfium/ictinus';
 
-<Overlay opem={true} onClose={() => {}}/>;
+<Overlay open={true} size={'33%'} onClose={() => {}}/>;
 ```
 
 # Props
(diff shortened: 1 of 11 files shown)


src/components/Overlay/Overlay.style.ts
```diff
@@ -0,0 +1,84 @@
...
+    case 'right':
+      return open ? 'translateX(0)' : 'translateX(100%)';
+    case 'bottom':
+      return open ? 'translateY(0)' : 'translateY(100%)';
+    default:
+      return open ? 'translateX(0)' : 'translateX(-100%)';
+  }
+};
+
+export const BackdropStyle = ({ open, anchor }: { open: boolean; anchor: AnchorType }) => (
```
Christos-Zacharopoulos, line 44: should be camelCase

Christos-Zacharopoulos: check other style functions as well

src/components/Overlay/Overlay.stories.mdx
```diff
@@ -0,0 +1,30 @@
...
+import { Overlay } from '@orfium/ictinus';
+
+<Overlay opem={true} onClose={() => {}}/>;
+```
+
+# Props
+
+<Props of={Overlay} />
+
+# Overlay
```
Christos-Zacharopoulos, line 24: Add a second story "Overlay Playground" and add knobs in order for the design team to see all the variations.

src/components/Overlay/Overlay.style.ts
```diff
@@ -0,0 +1,84 @@
...
+  theme: Theme
+): SerializedStyles => css`
+  ${flex};
+  position: fixed;
+  width: 100vw;
+  height: 100vh;
+  z-index: 3000;
```
Christos-Zacharopoulos, line 51: This has the same zIndex as the modal. Which one should be on top? 👀

Christos-Zacharopoulos (commented):
Very minor observations. Other than that looks good. Re-request review when you are done with these.

src/components/Overlay/Overlay.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+A universal Overlay component.
+
+# Usage
+
+```js
+import { Overlay } from '@orfium/ictinus';
+
+<Overlay opem={true} onClose={() => {}}/>;
```
mkarajohn, line 18: open

mkarajohn (commented):
Implementation wise this is fine, however I have some general thoughts on this component.

First off, the name. This fits the description of a drawer better than the existing `Drawer`. The existing `Drawer` better fits the description of side navigation or just navigation. "Overlay" tends to mean just the translucent backdrop, in my experience.

Secondly, since this is a component that is directly affected by screen size, as can be seen in ADR which uses something similar. It would be nice if there was a native way to define various widths according to screen size. Something like 

```typescript
type size = number | string | {
  lg: number|string; 
  md: number|string; 
  sm: number|string;
}
```
I would allow for string to give the user the ability to define the units, and not restrict to just `%`. i.e. `500px`

Thoughts @Orfium/front-end ?

Christos-Zacharopoulos:
> First off, the name. This fits the description of a drawer better than the existing `Drawer`. The existing `Drawer` better fits the description of side navigation or just navigation. "Overlay" tends to mean just the translucent backdrop, in my experience.
 

1) 
Completely agree about the `Drawer` component having a being poorly named. There is already a ticket in V5 to rename it to `Navigation` .

I don't think however that this is a bad name. `Overlay` is just a container with the minimum functionality of the onClose handler to run on the ClickAwayListener as well as the icon on the top left corner. It can be extended to cater for the needs of Modals as well.

> Secondly, since this is a component that is directly affected by screen size, as can be seen in ADR which uses something similar. It would be nice if there was a native way to define various widths according to screen size. Something like
> 
> ```ts
> type size = number | string | {
>   lg: number|string; 
>   md: number|string; 
>   sm: number|string;
> }
> ```
> 
> I would allow for string values as well to give the user the ability to define the units, and not restrict to just `%`. i.e. `500px`

The behavior of the component is the decision of the design team. I wouldn't mind having a fixed value , if this meets the needs of our use cases better. As far as I can remember, we had discussed this with @SVonKarajann and he had informed the design team that 33% of 1920 is different in width to 33% of 1200.

If we were to go with this 

```ts
type size = number | string | {
   lg: number|string; 
   md: number|string; 
   sm: number|string;
}
```

I would prefer the Keys of the object to be breakpoints. But still, I think we are over-thinking something that has already been researched and asked.

mkarajohn:
>I would prefer the Keys of the object to be breakpoints
The type I proposed was just for reference, the key point was to show a way to define various widths

Christos-Zacharopoulos:
> > The behavior of the component is the decision of the design team.
> 
> Can't we challenge the decision? It is obvious that, as is, this does not scale well to various screen sizes. If that's something that we do not care about at this point that's ok.

You skipped the part where I said we challenged them. 

>> As far as I can remember, we had discussed this with @SVonKarajann and he had informed the design team that 33% of 1920 is different in width to 33% of 1200.

I don't mind the type being ONLY a string and allow the user to pass whatever it is to width of the component.

It's just that then we need to add also a way to add a max and min width. It's not enough to define just the width if we go with a fixed value. 
 
> > It can be extended to cater for the needs of Modals as well.
> 
> I must be missing something in the implementation, because as is it is pretty hardwired to only serve as a drawer that comes out of the sides.

That's why I said extended.

SVonKarajann:
After discussing that with the DT as weell i changed the size type to string, so that the users can deside on the unit they wanna use

Christos-Zacharopoulos (approved):
LGTM! 🎉 
Kudos @mkarajohn also!

PR #444, opened 2022-01-17
[NO-JIRA] Refactor / palette functions

Christos-Zacharopoulos:
## Description

I started this an an investigation to check out if it's  doable. It turns out that it's feasible so I made it a PR.

I wanted to remove the `enhancePaletteWithShades` from the `pickTextColorFromSwatches` function since it's a recursive function and it's not something we have to run each and every time we need to have a AA text.

I also moved the getColor function to the palette since it's a palette function.

Finally, I renamed the `pickTextColorFromSwatches` to `getAAColorFromSwatches` and the `getAATextColor` to `getAAColor` and added them to theme.utils.

## Test Plan

Tested the getAAColor individually. No futher tests needed as it was more of a performance gain and no visual changes where made.

### DIFF
--- a/src/components/ButtonBase/ButtonBase.style.ts
+++ b/src/components/ButtonBase/ButtonBase.style.ts
@@ -2,7 +2,6 @@ import { transparentize } from 'polished';
 import { rem } from 'theme/utils';
 
 import { Theme } from '../../theme';
-import { pickTextColorFromSwatches } from '../../theme/palette';
 import { RequiredProperties } from '../../utils/common';
 import { ColorShapeFromComponent } from '../../utils/themeFunctions';
 import { Props } from './ButtonBase';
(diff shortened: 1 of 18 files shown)


Christos-Zacharopoulos:
image

src/components/ButtonBase/ButtonBase.style.ts
```diff
@@ -66,7 +65,7 @@ export const buttonBaseStyle = ({
       );
     }
 
-    return pickTextColorFromSwatches(calculatedColor.color, calculatedColor.shade);
+    return theme.utils.getAAColorFromSwatches(calculatedColor.color, calculatedColor.shade);
```
whosyourtaco, line 68: Can't you directly import `getAAColorFromSwatches`? There're many usages of it and having the `theme.utils` each time seems kinda redundant (?)

whosyourtaco (approved):
Nice job! LGTM

src/components/ButtonBase/ButtonBase.style.ts
Christos-Zacharopoulos: We can of course , but it's about being consistent in the way we use the theme utils. It's like importing the `getColor` function. We can do it, but we always prefer to use it through the theme.

PR #445, opened 2022-01-19
[DS-321]: fix(tooltip): add missing css props

SVonKarajann:
## Description
[Ticket](https://orfium.atlassian.net/browse/DS-321)


Christos-Zacharopoulos (approved):
image

PR #448, opened 2022-01-24
[DS-305] Feat / filter integration 

Christos-Zacharopoulos:
## Description
[Ticket ](https://orfium.atlassian.net/browse/DS-305)
Integration with Global states. 

This one has the exception that in pressed the global rules do not apply.

I haven't implemented some kind of config in this. It's something that I would like to further discuss and decide in the next FE chapter discussion in order to further implement in the future.

## Test Plan

Updated snaps

### DIFF
--- a/src/components/Filter/Filter.style.ts
+++ b/src/components/Filter/Filter.style.ts
@@ -2,10 +2,11 @@ import { css } from '@emotion/react';
 import { rem } from 'theme/utils';
 
 import { Theme } from '../../theme';
+import { getFocus, getHover, getPressed } from '../../theme/states';
 import { ButtonStyleProps } from './types';
 import {
   borderStyleParams,
-  focusBorderStyleParams,
+  focusBorderWidth,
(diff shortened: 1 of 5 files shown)


src/components/Filter/Filter.style.ts
```diff
@@ -43,25 +44,37 @@ export const buttonWrapperStyle = ({
...
           ? hasSelectedValue
-            ? theme.utils.getColor(calculatedColor.color, 100)
-            : theme.utils.getColor('darkGrey', null, 'pale')
+            ? getHover({ theme, color: calculatedColor.color, shade: calculatedColor.shade })
+                .backgroundColor
+            : getHover({ theme }).backgroundColor
+          : undefined,
+    },
+
+    ':active > div': {
```
panvourtsis, line 57: this is difficult to capture can we have a section with comments explain the logic ?

panvourtsis (approved):

PR #451, opened 2022-02-02
[DS-324] Fix/ select auto scroll bug

Christos-Zacharopoulos:
## Description

Bug fix on scroll. The ref was mounted to a wrong div AS WELL as the right one. So the wrong div was overriding the correct one and it was scrolling there.

Ticket: https://orfium.atlassian.net/browse/DS-324


panvourtsis:
Can you explain this a bit. I don't understand how this `ref` cause the auto scroll use

Christos-Zacharopoulos:
> Can you explain this a bit. I don't understand how this `ref` cause the auto scroll use

Sure. We are passing a `ref` for the selected item and we are using this `ref` to auto scroll on open to the correct item. 

The `ref` was attached to the `ListItem` component if it was selected BUT it was also wrongly attached to all the `ListItemGroupTitles`. The last `ListItemGroupTitle` is rendered after the render of the selected `ListItem` so it was overriding the `ref`

PR #452, opened 2022-02-07
[DS-297]: feat(textInputBase): testInputBase updated

SVonKarajann:
## Description
[Ticket](https://orfium.atlassian.net/browse/DS-297)

Currently in DS TextInputBase component is used by, TextField component (which is also used by Select component), SearchField component and TextArea component.
Untitled Diagram drawio

Since all of these components shared the exact same css rules, with a few differences in some cases, i moved those rules to `TextInputBase.style.ts` file. 

Also so far we determined the size of the TextInputBase by adding different top and bottom padding for each size case. I changed that by setting  the display of the wrapperStyle to flex, with align-items rule set to 'center', and adding the exact height defined by the DT.

Also fixed a few bugs
areatext-bug
input-bug

fixes:
areatext-fix
input-fix


SVonKarajann:
@Orfium/front-end Kind reminder

Christos-Zacharopoulos (approved):
Awesome work!

PR #456, opened 2022-02-17
[DS-336] Feat/ Error handling 

Christos-Zacharopoulos:
## Description

In this PR we implement a way to check and throw errors based on Props passed to component.

Implementation decisions taken:

When starting, I thought that using a HOC would be the optimum way to handle this case.While analyzing this I soon figured out that using a HOC was not possible. 

We are and should be defining default prop values on components. These values have not initialized while in the HOC scope so it's impossible to make business checks there.

So I tried using a function to take props and check them.  Problem is that this made everything more complicated hindered readability as the props where de-structured in order to apply default values and they were used again in the function.

image

So finally, by abstracting the default props to their own object and passing them to the component, as seen in Reacts' official [documentation](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values), the component shows the correct values in storybook props and can make complex prop based calculations to check if there is an error and throw the appropriate message.

**_Updates to initial comment and implementation:_**
https://github.com/Orfium/orfium-ictinus/pull/456#issuecomment-1062908974

ticket: https://orfium.atlassian.net/browse/DS-336

## Test Plan

Test cases covered:
- at least one condition is met and error is thrown
- no condition is met and error is not thrown


mkarajohn:
I have some thoughts on this, which I will list in no particular order:

>So finally, by abstracting the default props to their own object and passing them to the component, as seen in Reacts' official [documentation](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values),

When I read that I immediately remembered [this](https://github.com/facebook/react/pull/16210) and [this](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components). 
tl;dr default props are going to be deprecated at some point

> the component shows the correct values in storybook props 

There was a [similar issue](https://github.com/storybookjs/storybook/issues/9626) in the storybook repo, but that was for v5 of storybook. According to the docs this functionality is supposed to be implemented in v6+ (we are using 6.3.5 as of the point of writing this). Relevant links: 
* https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/props-tables.md#customizing-argtypes
* https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#writing-stories
* https://storybook.js.org/docs/react/essentials/controls

We can research these together if you want and see if the cover our need for correct default props in the Storybook docs.

> can make complex prop based calculations to check if there is an error and throw the appropriate message.

While what you have prototyped here definitely works, could the same effect be achieved solely with typescript? I went ahead and recreated the cases of the erroneous props described in the [1st error condition](https://github.com/Orfium/orfium-ictinus/pull/456/files#diff-35dda77773b93791bdad4f279204342cefd6803a36288d6ad2dff1df308a07b3R10-R15) and it seems to be working, alas without the descriptive error message, but does the error message matter if a) the TS compiler throws and b) it's documented in the docs which properties are incompatible with which properties? e.g. the description of the `isSelected` prop could read something like "Can only be `true` if the `styleType` is set to `interactive`" or "Values are ignored if `styleType` is set to `readonly`".

In conclusion, the overall theme of my thoughts is this: Are we over-engineering this?

Christos-Zacharopoulos:
@mkarajohn First of all thank you for dedicating the time and effort on this.

Answering straight to the conclusion as the rest is another discussion about technical implementation.

The point of this is not to just resolve some issues that may occur in runtime. As you said, these can be documented thoroughly and try to avoid these issues. 

Ictinus is just another tool for us to develop products in Orfium. And as with every tool and project, we need to always find ways to make the code more readable and easier to maintain. 

That's the reason behind this change. To have a common way of handling whatever kind of errors might arise from an object. To find a clean way of abstracting the error cases from the body of a function but still keep use the same way to handle all cases inside Ictinus.

This can be used in plain Javascript as well, so functions like getColor won't have to have their body _polluted_ with conditions and `throw new Error`. 

I don't think I am over-engineering anything, but hardly ever do those that do it understand it immediately. 😅

src/utils/helpers.ts
```diff
@@ -33,7 +33,7 @@ export const getLocaleFormat = (dateFormat: string | undefined) => {
  *  A function that takes an array of errors and the component props and throws an error
  *  if the condition is met
  * */
-export const propsHandler = <T>(
+export const errorHandler = <T>(
   errors: { condition: (p: T) => boolean; message: string }[],
   props: T
```
mkarajohn, line 38: Maybe we could add a 3rd optional argument with the exact error we want to have. For example

```js
class PropsValidationError extends Error {
  constructor(message) {
    super(message);
    this.name = "PropsValidationError"; 
  }
}
```

```js
...
errorHandler(validators, props, PropsValidationError);
...
```

mkarajohn (approved):

src/utils/helpers.ts
Christos-Zacharopoulos: Great idea on the custom error but in order for this to be cleaner and extendable wouldn't it be better to pass the custom Error instead of the message and just throw it?

mkarajohn: Unless i am not understandng the question, this is what my example illustrates. You create a custom `Error`, you pass it as a 3rd optional argument, and whatever throws from these validators will be a custom `Error` :thinking:

Christos-Zacharopoulos: Check the last commit to understand what I meant

mkarajohn: Ah, I see. My original suggestion was based on the assumption that a given group of validators would fall into a specific "error category" so all exceptions would throw the same kind of error.

Christos-Zacharopoulos: Yes, but I thought it's better to allow for different errors.

This way, the same component could throw different errors and we could have different handling for each error in the future.    

Also I didn't want to add a third argument in the function to keep it simpler and easier to test.

Christos-Zacharopoulos:
image

Christos-Zacharopoulos:
image

PR #461, opened 2022-02-22
[NO-JIRA] fix/ci workflow for docs creation

Christos-Zacharopoulos:
## Description

This PR changes the target folder used to copy the generated docs from. Here we already use the filename docs for other purposes.


panvourtsis (approved):
nice one

PR #464, opened 2022-02-23
[DS-340] fix/button refactor to reduce cognitive complexity

Christos-Zacharopoulos:
## Description

This is a first attempt to refactor a function labeled for high [cognitive complexity](https://www.sonarsource.com/resources/white-papers/cognitive-complexity/#:~:text=Cognitive%20Complexity%20offers%20a%20new,strikes%20developers%20as%20intuitively%20right.&text=Cognitive%20Complexity%20breaks%20from%20the,models%20to%20assess%20software%20maintainability.). Analysis runs on master releases only for now.

Ticket: https://orfium.atlassian.net/browse/DS-340

### DIFF
--- a/src/components/ButtonBase/ButtonBase.style.ts
+++ b/src/components/ButtonBase/ButtonBase.style.ts
@@ -6,7 +6,7 @@ import { RequiredProperties } from '../../utils/common';
 import { ColorShapeFromComponent } from '../../utils/themeFunctions';
 import { Props } from './ButtonBase';
 import { buttonConfig, buttonSizes } from './config';
-import { defineBackgroundColor } from './utils';
+import { calculateButtonColor, defineBackgroundColor } from './utils';
 
 /** Calculates the button specific height based on the size passed to it **/
 export const heightBasedOnSize = (size: typeof buttonSizes[number]) =>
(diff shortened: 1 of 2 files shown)


src/components/ButtonBase/ButtonBase.style.ts
```diff
@@ -33,29 +33,22 @@ export const buttonBaseStyle = ({
...
-      return theme.utils.getColor(color.name, color.shade);
-    }
-
-    if (isOutlined || transparent) {
-      return backGroundColor;
-    }
-
-    return theme.utils.getAAColorFromSwatches(calculatedColor.color, calculatedColor.shade);
-  };
+  const isBackgroundTransparent = isOutlined || transparent;
```
mkarajohn, line 36: I found these 2 lines are confusing. I mean, in the end I get what is happening, but I actually had to think about what is going on.

mkarajohn (approved):

src/components/ButtonBase/ButtonBase.style.ts
Christos-Zacharopoulos: I agree. This is something that will be partly fixed on v5 where the type of the button will be one variable instead of relying on 2 bool.

Christos-Zacharopoulos: v5 ticket [here](https://orfium.atlassian.net/browse/DS-326)

PR #465, opened 2022-02-25
[DS-343] - Broadcast Icon

kostasdano:
## Description

[DS-343](https://orfium.atlassian.net/browse/DS-343)


panvourtsis (approved):
LGTM 🎊

PR #466, opened 2022-03-09
[DS-347] fix/Checkbox positioning bug

mkarajohn:
## Description

The checkbox input element is absolutely positioned within a span without an explicit position value. This causes issues when there are other positioned ancestors higher up in the DOM tree.

## Screenshot
Peek 2022-03-10 00-53


SVonKarajann (approved):

panvourtsis (approved):
so simple yet so powerful 🎊

panvourtsis:
I am merging this since there is no need for more QA as the fix and UI is way to obvious

PR #474, opened 2022-03-18
[DS-348]: feat: align all textfield based components with ds width rules

whosyourtaco:
## Description

- Replaces min-width with the DS value (150px)
- Removes max-width from the Select component
- Enables full fluidity of TextInputBase components, according to the parent div's width

### DIFF
--- a/src/components/Select/Select.style.ts
+++ b/src/components/Select/Select.style.ts
@@ -9,7 +9,6 @@ export const selectWrapper = ({
 }) => (): SerializedStyles => css`
   position: relative;
   min-width: ${rem(150)};
-  max-width: ${rem(620)};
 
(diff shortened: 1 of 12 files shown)


src/components/Select/Select.style.ts
```diff
@@ -9,7 +9,6 @@ export const selectWrapper = ({
 }) => (): SerializedStyles => css`
   position: relative;
   min-width: ${rem(150)};
-  max-width: ${rem(620)};
```
panvourtsis, line 12: @kostasdano this will affect the large selects of CMO you have to do it on the product now

panvourtsis (approved):
LGTM 🎊

PR #481, opened 2022-03-30
feat: stabilize dark mode and add new dark palette 3/3

panvourtsis:
## Description

depends on https://github.com/Orfium/orfium-ictinus/pull/480

*This PR will be a part of 3 that has a goal for a small fix that I made for dark mode first try. 3/3*

This PR introduces a new dark mode config based on Figma and extends the current functions to apply a basic dark mode functionality to the app.

### TextField
Screen Recording 2022-03-30 at 10 59 06 AM

### Button
Screen Recording 2022-03-30 at 10 58 44 AM

### Checkbox
Screen Recording 2022-03-30 at 10 58 26 AM

## Test Plan

* Update state test because I changed the `darkBlue` to `blue`
* introduce `ThemeProvider.test.tsx` to test the dark mode switching

### DIFF
--- a/src/components/Radio/Radio.style.ts
+++ b/src/components/Radio/Radio.style.ts
@@ -9,13 +9,7 @@ import { Props } from './Radio';
 const lightHoverColor = 'rgba(14, 14, 23, 0.07)';
 const darkHoverColor = 'rgba(255, 255, 255, 0.1)';
 
-const boxShadow = ({
-  inset = false,
-  colorScheme,
-}: {
-  inset?: boolean;
-  colorScheme: ColorScheme;
-}) => css`
+const boxShadow = ({ colorScheme }: { colorScheme: ColorScheme }) => css`
   box-shadow: 0 0 0 ${rem('6px')} ${colorScheme === 'dark' ? darkHoverColor : lightHoverColor},
     inset 0 0 0 ${rem('6px')} ${colorScheme === 'dark' ? darkHoverColor : lightHoverColor};
 `;
@@ -45,7 +39,7 @@ export const customRadioInnerHover = (focused: boolean, disabled: boolean) => (
   width: ${rem('24px')};
   height: ${rem('24px')};
   transition: all 0.2s ease;
-  // ${focused && !disabled && boxShadow({ inset: true, colorScheme: theme.colorScheme })};
+  ${focused && !disabled && boxShadow({ colorScheme: theme.colorScheme })};
 `;
 
 export const customRadioWrapperStyles = (focused: boolean, disabled: boolean) => (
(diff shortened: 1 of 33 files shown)


elektracodes (approved):

mkarajohn (changes requested):
Please rebase on 480

mkarajohn:
Without yet checking the code, shouldn't the icons turn black as well when dark mode is enabled? White icons on this background is not easily readable
Screen Recording 2022-03-30 at 10 58 44 AM

src/components/Radio/Radio.style.ts
```diff
@@ -1,14 +1,23 @@
...
 import { Props } from './Radio';
 
-const hoverColor = 'rgba(0, 0, 0, 0.05)';
+const lightHoverColor = 'rgba(14, 14, 23, 0.07)';
+const darkHoverColor = 'rgba(255, 255, 255, 0.1)';
 
-const focusedRadio = css`
-  box-shadow: 0 0 0 ${rem('12px')} ${hoverColor};
+const boxShadow = ({
+  inset = false,
```
SVonKarajann, line 13: `inset` param in never used

src/components/Radio/Radio.style.ts
```diff
@@ -80,8 +86,33 @@ export const customRadioStyles = (props: Pick<Props, 'checked' | 'disabled' | 'f
     width: 100%;
     height: 100%;
     box-sizing: border-box;
-    box-shadow: ${determineBoxShadow(props)(theme)};
     position: absolute;
+    // background-color: ${determineColorBasedOnState(props)(theme)};
```
SVonKarajann, line 90: Remove if not necessary?

src/components/TextInputBase/config.ts
```diff
@@ -11,42 +11,70 @@ export const MIN_WIDTH = 150;
...
+        border: {
+          width: 1,
+          color: {
+            default: {
+              name: 'darkGrey',
+              shade: 500,
+            },
+            pressed: {
+              name: 'blue',
+              shade: 350,
```
SVonKarajann, line 69: Is there any rule given from the DT on how we change shade values from light to dark mode?

mkarajohn (commented):
Not approving because of https://github.com/Orfium/orfium-ictinus/pull/481#issuecomment-1084456291. Otherwise seems ok

panvourtsis:
> Without yet checking the code, shouldn't the icons turn black as well when dark mode is enabled? White icons on this background are not easily readable Screen Recording 2022-03-30 at 10 58 44 AM

For this, there is no use of icons inside the component. There is a plan to replace this in the next versions. Until then the icons are passed as new components to the button so will be wrong to overwrite styles

src/components/TextInputBase/config.ts
panvourtsis: whatever they have on the Figma rules on 4.5. This is not a final dark mode functionality but a very basic demo of how it works based on the DT specifications. Check the Figma for these

mkarajohn (approved):

PR #483, opened 2022-04-08
POC / css prop

Christos-Zacharopoulos:
## Description

POC on css prop

- add sx prop to TextInputBase and integrate textArea

Also, fixed a small issue on TextAree
Wrong:
snipWrong
Fixed:
snip

### DIFF
--- a/src/components/TextArea/TextArea.style.ts
+++ b/src/components/TextArea/TextArea.style.ts
@@ -1,13 +1,18 @@
-import { css, SerializedStyles } from '@emotion/react';
+import { CSSObject } from '@emotion/serialize';
 
 import { Theme } from '../../theme';
-import { Props } from './TextArea';
-import { Props as TextInputWrapperProps } from 'components/TextInputBase';
-import { inputStyle as baseInputStyle } from 'components/TextInputBase/TextInputBase.style';
+import { rem } from '../../theme/utils';
+import { MIN_WIDTH } from '../TextInputBase/config';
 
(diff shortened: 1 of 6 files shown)


Christos-Zacharopoulos:
@Orfium/front-end Do we like this kind of approach? Should I move forward with it?

whosyourtaco (commented):
.

whosyourtaco:
@Christos-Zacharopoulos This is really nice! It will really improve our workflow. 
I only have one note. Adding/replacing custom properties in our components is discouraged both from the DS team and our team. Still, situations arise that custom properties are requested. By implementing this property in this way, we are allowing the dev to overwrite any property easily. Maybe we could decide which properties should be allowed to be overwritten and which ones shouldn't, and then move the spread operator accordingly?

This is really minor, I just feel that if we make it easier to overwrite styles, it will soon become a norm

Christos-Zacharopoulos:
> Maybe we could decide which properties should be allowed to be overwritten and which ones shouldn't, and then move the spread operator accordingly?
>

This was created after [this](https://orfium.atlassian.net/wiki/spaces/FE/pages/2420867249/RFC+-+Configuration+files+Style+prop) RFC. 

The sx prop will be used for now only internally exactly to 
`Enable the centralized components to be even cleaner and more abstracted `

We could start using it in the future but it would be even more important to find what we want to allow a user to do and what not.

Finally, 
> we are allowing the dev to overwrite any property easily
this is something that will create even more custom cases in our products instead of additions to the DS. We could discuss this further with the rest of the team and decide the way forward on this.

whosyourtaco (approved):

PR #488, opened 2022-04-27
[DS-366] feat:Update implementation of NotificationsContainer.tsx to be able to be placed within other elements

mkarajohn:
## Description

Resolves [DS-366]

## Test Plan

[DS-366]: https://orfium.atlassian.net/browse/DS-366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mkarajohn:
Yes, that's correct. If an element is not found by one of the usual DOM
methods (`querySelector()` etc) or even by React refs, the returned value
is `null`

whosyourtaco (approved):

PR #364, opened 2021-09-21
[WIP] Feat/figma integration

panvourtsis:

PR #493, opened 2022-05-18
[DS-319] Feat/ Figma intefration

Christos-Zacharopoulos:
## Description

Ticket: [DS-319](https://orfium.atlassian.net/browse/DS-319)
https://github.com/Orfium/orfium-ictinus/pull/364
> This PR introduces a new integration for Figma. Now in each case, we can see the exact frame of the component in Figma and compare.
> 
> This will improve performance while we develop and also will provide a live comparison on both dev and production environment

## Screenshot

image


kostasdano (approved):
approve

PR #496, opened 2022-06-02
[DS-377]: feat: adds Slider component

whosyourtaco:
## Description

- Adds the Slider component, as shown in Figma

Can be considered a WIP, as it is blocked by [DS-382], which has an issue with the text field sizes. Will update this PR when ready.

### DIFF
--- a/src/components/Slider/Slider.tsx
+++ b/src/components/Slider/Slider.tsx
@@ -1,0 +1,101 @@
+import React, { useEffect } from 'react';
+import { Range } from 'react-range';
+
+import { TestProps } from '../../utils/types';
+import SliderMark from './components/SliderMark';
+import SliderThumb from './components/SliderThumb';
+import SliderTrack from './components/SliderTrack';
+import { Container } from './Slider.style';
+
+interface Props {
+  /** Determines the position of the rendered thumbs. Defaults to [0, 100] */
+  values: number[];
+  /** Determines if the Slider is disabled or not */
+  disabled?: boolean;
+  /** Determines whether the Slider is a Range or a Selector */
+  isSelector?: boolean;
+  /** Determines if the Slider will show increments along its track */
+  hasIncrements?: boolean;
+
+  /** Called when a thumb is moved, provides new values */
+  onChange: (values: number[]) => void;
+  /** Called when a change is finished (mouse/touch up, or keyup), provides current values.
+   * Use this event when you have to make for example ajax request with new values. */
+  onFinalChange?: (values: number[]) => void;
+}
+
+const STEP = 1;
+const STEP_WITH_INCREMENTS = 20;
+export const MIN = 0;
+export const MAX = 100;
+
+const Slider: React.FC<Props & TestProps> = ({
+  values,
+  onChange,
+  onFinalChange,
+  isSelector = false,
+  hasIncrements = false,
+  disabled = false,
+  dataTestPrefixId,
+}) => {
+  useEffect(() => {
+    if (isSelector && values.length > 1) {
+      throw new Error(
+        'The Selector type Slider can only accept one initial value in the array, since it only has one thumb. Try changing the `initialValues` prop.'
+      );
+    }
+  }, [values, isSelector]);
+
+  return (
+    <Container data-testid={`${dataTestPrefixId ?? ''}slider_component`}>
+      <Range
+        step={hasIncrements ? STEP_WITH_INCREMENTS : STEP}
+        min={MIN}
+        max={MAX}
+        disabled={disabled}
+        values={values}
+        onChange={onChange}
+        onFinalChange={onFinalChange}
+        renderMark={({ props, index }) =>
+          hasIncrements &&
+          (isSelector ? index > MIN : index > MIN && index < MAX / STEP_WITH_INCREMENTS) && (
+            <SliderMark
+              key={`mark_${index}`}
+              dataTestPrefixId={dataTestPrefixId}
+              restProps={props}
+              values={values}
+              index={index}
+              isSelector={isSelector}
+              disabled={disabled}
+            />
+          )
+        }
+        renderTrack={({ props, children }) => (
+          <SliderTrack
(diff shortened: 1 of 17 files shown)


panvourtsis:
@whosyourtaco reminder

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+import useTheme from '../../hooks/useTheme';
+import { TestProps } from '../../utils/types';
+import TextField from '../TextField';
+import SliderMark from './components/SliderMark';
+import SliderThumb from './components/SliderThumb';
+import SliderTrack from './components/SliderTrack';
+import { Container, InputContainer, InputsContainer } from './Slider.style';
+
+interface Props {
```
mkarajohn, line 12: If `Selector` only has 1 value and `Range` only has 2 values, the props could be split as 

```ts
type CommonProps = {
  disabled?: boolean;
  hasIncrements?: boolean;
  onChange: (values: number[]) => void;
  onFinalChange?: (values: number[]) => void;
}

type SelectorProps = {
  values: [number];
  isSelector: true;
  initialValues?: [number];
};

type RangeProps = {
  values: [number, number];
  isSelector: false;
  initialValues?: [number, number];
};

type Props = CommonProps & (SelectorProps | RangeProps);
```

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+const Slider: React.FC<Props & TestProps> = ({
+  values,
+  initialValues = [MIN, MAX],
+  onChange,
+  onFinalChange,
+  isSelector = false,
+  hasIncrements = false,
+  disabled = false,
+  dataTestPrefixId,
```
mkarajohn, line 43: This is not declared in the props type. Also, IMO, it would make more sense to call it `dataTestIdPrefix`

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,157 @@
...
+        step={hasIncrements ? STEP_WITH_INCREMENTS : STEP}
+        min={MIN}
+        max={MAX}
+        disabled={disabled}
+        values={values}
+        onChange={onChange}
+        onFinalChange={onFinalChange}
+        renderMark={({ props, index }) =>
+          hasIncrements &&
+          (isSelector ? index > MIN : index > MIN && index < MAX / STEP_WITH_INCREMENTS) && (
```
mkarajohn, line 79: Maybe replace this whole ternary operation with a single variable with a descriptive name?

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,157 @@
...
+    <Container data-testid={`${dataTestPrefixId ?? ''}slider_component`}>
+      <Range
+        step={hasIncrements ? STEP_WITH_INCREMENTS : STEP}
+        min={MIN}
+        max={MAX}
+        disabled={disabled}
+        values={values}
+        onChange={onChange}
+        onFinalChange={onFinalChange}
+        renderMark={({ props, index }) =>
```
mkarajohn, line 77: With regards to the comment below, I personally always prefer writing a proper function body, instead of the return shorthand, because it actually allows you to do function things inside the function body, and helps with debugging too. 

I am just throwing this here, not requesting a change (even though if you decide to implement the change suggested in the  comment below, you will have to make this change here as well :hankey: )

mkarajohn (changes requested):

src/components/Slider/Slider.tsx
whosyourtaco: This is from the `TestProps` type that we import in ictinus' components in order to not declare a `dataTestId` prop every time :D

src/components/Slider/Slider.tsx
```diff
@@ -8,46 +8,65 @@ import SliderMark from './components/SliderMark';
...
-  isSelector = false,
   hasIncrements = false,
   disabled = false,
   dataTestPrefixId,
 }) => {
   const theme = useTheme();
 
+  const isSelector = useMemo(() => values.length === 1, []);
+
+  const initialValueRef = useRef<number[]>();
```
mkarajohn, line 54: `[number] | [number, number]` if you want to be more strict

src/components/Slider/Slider.tsx
```diff
@@ -8,46 +8,65 @@ import SliderMark from './components/SliderMark';
...
+  /** Determines the position of the rendered thumbs and the type of the Slider. 1 value means it's a Selector,
+   * while 2 values mean it's a Range. Defaults to [0, 100] */
+  values: [number];
+};
+
+type RangeProps = {
+  /** Determines the position of the rendered thumbs and the type of the Slider. 1 value means it's a Selector,
+   * while 2 values mean it's a Range. Defaults to [0, 100] */
+  values: [number, number];
+};
```
mkarajohn, line 35: Since this has been simplified, maybe now you could just have a single `Props` with a 

```
/** Determines the position of the rendered thumbs and the type of the Slider. 1 value means it's a Selector,
   * while 2 values mean it's a Range. Defaults to [0, 100] */
values: [number] | [number, number];
```

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,198 @@
...
+  dataTestPrefixId,
+}) => {
+  const theme = useTheme();
+
+  const isSelector = useMemo(() => values.length === 1, []);
+
+  const initialValueRef = useRef<[number] | [number, number]>();
+
+  useEffect(() => {
+    initialValueRef.current = values;
```
mkarajohn, line 48: ...I know I suggested it this way, but I just realized this can be written even simpler:

```ts
const initialValueRef = useRef<[number] | [number, number]>(props.values);
```

This will only run once, so you don't even need the `useEffect`, and then you can instantly set 

```ts
const isSelector = initialValueRef.current.length === 1;
```

mkarajohn (approved):

PR #497, opened 2022-06-06
feat(list): add support for icons

panvourtsis:
## Description

In this PR I added support for icons in the options of the list, and select items.

I added the support of `iconProps` where we directly use the props of the icon to define everything we want on the component. This way we have the flexibility of changing the list in anything we want

## Screenshot

Screenshot 2022-06-06 at 4 30 12 PM

## Test Plan
only snapshots

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -6,7 +6,8 @@ import useCombinedRefs from '../../hooks/useCombinedRefs';
 import useTheme from '../../hooks/useTheme';
 import { ChangeEvent } from '../../utils/common';
 import { TestProps } from '../../utils/types';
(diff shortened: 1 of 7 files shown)


src/components/Select/Select.tsx
```diff
@@ -15,13 +16,17 @@ import SelectMenu from './components/SelectMenu/SelectMenu';
 import { rightIconContainer, selectWrapper } from './Select.style';
 import Loader from 'components/Loader';
 
-export type SelectOption = {
```
panvourtsis, line 18: I changed this part so we can have only one reference of what an Option should be like. So this will be used by anyone else

Christos-Zacharopoulos (approved):

PR #501, opened 2022-06-15
[NDS-335]: fix: fixes drawer styles between states

whosyourtaco:
## Description

- Fixes an issue where the hover and pressed effects didn't show when the item had no submenus
- Fixes issue where the icon was displayed wrong when the item had no submenus

PS: The styling in this component is very off and needs to be updated. This is just a small PR to fix a minor issue for the time being

### DIFF
--- a/src/components/Drawer/Navigation/Navigation.style.ts
+++ b/src/components/Drawer/Navigation/Navigation.style.ts
@@ -5,9 +5,11 @@ import { BASE_SHADE } from 'theme/palette';
 import { rem } from 'theme/utils';
 import { fillPickerBasedOnType } from 'utils/themeFunctions';
 
-import { getFocus, getHover, getPressed } from '../../../theme/states';
+import { getFocus, getHover } from '../../../theme/states';
 
 const ICON_PADDING = 39;
+
+//TODO The whole logic and styling here is off compared to the designs and needs to be changed in the next version.
 
 export const navigationContainerStyle = (expanded: boolean) => (): SerializedStyles => css`
   ${transition(10.2)};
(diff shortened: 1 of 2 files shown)


src/components/Drawer/Navigation/Navigation.style.ts
```diff
@@ -32,8 +32,31 @@ const itemStyle = (theme: Theme): SerializedStyles => css`
   cursor: default;
 `;
 
+const menuStateStyles = (theme: Theme): SerializedStyles => css`
+  &:hover {
+    background-color: ${getHover({ theme }).backgroundColor}; !important;
```
mkarajohn, line 37: Can you explain the logic of background-coloring here? I think that since every state has `!important` and it's not used to add arbitrary specificity to one state over the others, then it could be omitted altogether?

mkarajohn (commented):
Just saw that the code over which I commended on, was just moved there and preexisted, but if it in fact is not necessary to have all those `!important` maybe we could remove them

whosyourtaco:
> Just saw that the code over which I commended on, was just moved there and preexisted, but if it in fact is not necessary to have all those `!important` maybe we could remove them



> Just saw that the code over which I commended on, was just moved there and preexisted, but if it in fact is not necessary to have all those `!important` maybe we could remove them

Yeah it was already implemented this way, but, as it turns out, they weren't necessary after all, nice catch!

mkarajohn (approved):

whosyourtaco:
Turns out the designs were wrong and these changes are in turn wrong. This will be frozen until v5 clears up the Drawer component and we can implement the new one correctly.

PR #505, opened 2022-06-22
 sonar workflow running correctly

panvourtsis:
## Description

After investigating our sonarqube on ictinus I found out that we don't have the coverage passed to it. This PR fix that and I replaced the `push` with `pull_request` as a suggestion to run in every PR.

We expect to have 85% coverage after merging this to the master.

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -1,4 +1,4 @@
-name: Build
+name: SonarQube
 on:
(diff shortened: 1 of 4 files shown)


Christos-Zacharopoulos (approved):
image

Christos-Zacharopoulos:
Maybe add this to sonar documentation as well?

PR #506, opened 2022-06-30
fix: fix slider export and onchange method issue

whosyourtaco:
## Description

- Fixes the default export of the Slider component
- Fixes an issue where changing the TextFields didn't call the onFinalChange callback


Christos-Zacharopoulos (approved):

whosyourtaco:
After talking with @Christos-Zacharopoulos and @mkarajohn, there's a debate on whether the `onChange` and `onFinalChange` callbacks provided are a) utilized correctly and b) have correct naming.

Leaving this here so we can track it and revisit it in the future.

PR #509, opened 2022-07-07
[NDS-338] Feat/Props Standardization (Booleans pt1)

Christos-Zacharopoulos:
## Description

**FOR V5**

This PR is part of a chain of PRs to follow regarding props globalization.

In particular this PR introduces an eslint rule to enforce the use of the prefix `is` on boolean props to our components.

While making this change, it was understood that in Modal the prop `closeOnEsc` is not needed and this should be the behavior of the component. 

ticket: https://orfium.atlassian.net/browse/NDS-338

## Test Plan

Apart from Modal no other functionality has changed. Just updated snapshots where needed.

In modal removed the test case where `closeOnEsc` was tested.


panvourtsis (approved):

.eslintrc.json
```diff
@@ -2,18 +2,12 @@
   "extends": [
```
panvourtsis, line 2: only comments is here that we can use these to our projects as well.

.eslintrc.json
```diff
@@ -36,6 +30,13 @@
     "react/button-has-type": "warn",
     "react/no-array-index-key": "warn",
     "react/no-render-return-value": 0,
+    "react/boolean-prop-naming": [
+      "error",
+      {
+        "rule": "^(is)[A-Z]([A-Za-z0-9]?)+",
+        "message": "It is better if your prop ({{ propName }}) starts with `is` or `has`"
```
mkarajohn, line 37: "It is better if" does not imply error :eyes: 
Should be stronger, like "Your prop .... should start blah blah"

mkarajohn (approved):

PR #511, opened 2022-07-12
[NDS-341]: fix: rename onFinalChange to onBlur

whosyourtaco:
## Description

- As decided in our last FE Discussions, the `onFinalChange` prop is renamed to `onBlur`


mkarajohn (approved):
image

PR #542, opened 2022-11-29
[NDS-490]: theme reconfiguration spacing

kostasdano:
## Description

[NDS-490](https://orfium.atlassian.net/browse/NDS-490) 

[part I of [NDS-489](https://orfium.atlassian.net/browse/NDS-489)]

In this PR the **new spacing configuration** is introduced and all spacing usages are changed.

- Instead of **xsm, sm, md, lg, xl** we now introduce **spacings with number labels (0 to 12)**.
     - e.g.: `theme.spacing.md` becomes `theme.spacing.get('6') `
- We only have one **get** function (`theme.spacing.get`).
- In order to be aligned with the new spacing config, **Card** props types were changed (**radius**)


src/components/Menu/Menu.style.ts
```diff
@@ -4,22 +4,39 @@ import { Theme } from '../../theme';
...
-    marginRight: hasChildren ? theme.spacing.sm : 0,
-  },
-});
+const getButtonMargin =
+  ({ iconLeft, iconRight, size }: Pick<Props, 'iconLeft' | 'iconRight' | 'size'>) =>
+  (theme: Theme) => {
+    if (iconLeft || iconRight) {
+      if (size === 'sm') {
+        return theme.spacing.get('4');
+      } else {
```
panvourtsis, line 19: you dont really need else here

src/components/Menu/Menu.style.ts
```diff
@@ -4,22 +4,39 @@ import { Theme } from '../../theme';
...
+        return theme.spacing.get('6');
+      }
+    } else {
```
panvourtsis, line 22: and here

src/theme/utils.ts
```diff
@@ -82,3 +74,13 @@ export const getColorErrors = [
     error: new PropsValidationError('No color found with that name'),
   },
 ];
+
+/**
+ * 
+ * @param figmaTokensObject The parsed objects from Figma Tokens in the src/theme/constants/ dir
+ * @returns the value of the figma token item converted into rem
+ */
+export const getFigmaTokensValue =
```
panvourtsis, line 83: this seems as covered but in the case of and object with `val` undefined this could fail. It is covered by `rem` function that always returns string

case: `Number({}['test'])`

In the case above if you pass it in the `rem` function you will have `NaNrem` which is a non readable variable. 

We are covered by the fact that the application will not break but UI might. Maybe we can put a test here for future use? What do you think?

kostasdano: get by lodash seems pretty safe and optimal to me. maybe with a default value of 0(px).

panvourtsis (approved):

PR #544, opened 2022-12-02
[NDS-495] feat: change the first portion of components

panvourtsis:
## Description

This is part ONE of the goal, to fix all internal types so they can be easily accessible on the root of the library by an external user.

### DIFF
--- a/src/components/Chip/Chip.types.ts
+++ b/src/components/Chip/Chip.types.ts
@@ -1,6 +1,6 @@
 import { ClickEvent } from '../../hooks/useLoading';
 import { flatColors } from '../../theme/palette';
-import { ButtonProps } from '../../utils/common';
+import { CommonButtonProps } from '../../utils/common';
 import { TestProps } from '../../utils/types';
 
 export const READ_ONLY = 'read-only' as const;
@@ -8,7 +8,7 @@ export const INTERACTIVE = 'interactive' as const;
 
 export const styleType = [READ_ONLY, INTERACTIVE] as const;
 
-export type Props = {
+export type ChipBaseProps = {
   /**
    * Determines whether the chip should be read-only or interactive.
    * @default read-only
@@ -32,4 +32,4 @@ export type Props = {
   isDisabled?: boolean;
 };
 
-export type ChipProps = Props & TestProps & ButtonProps;
+export type ChipProps = ChipBaseProps & TestProps & CommonButtonProps;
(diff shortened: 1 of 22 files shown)


src/components/Chip/Chip.types.ts
```diff
@@ -1,14 +1,14 @@
 import { ClickEvent } from '../../hooks/useLoading';
 import { flatColors } from '../../theme/palette';
```
kostasdano, line 2: GitHub Actions Failed Workflow indicates an import error

`Module '"../../utils/common"' has no exported member 'CommonButtonProps'.`

src/components/Chip/Chip.types.ts
```diff
@@ -1,14 +1,14 @@
...
-import { ButtonProps } from '../../utils/common';
+import { CommonButtonProps } from '../../utils/common';
```
kostasdano, line 3: GitHub Actions Failed Workflow indicates an import error

`Module '"../../utils/common"' has no exported member 'CommonButtonProps'.`

src/components/Chip/components/Badge/Badge.tsx
```diff
@@ -2,14 +2,14 @@ import React from 'react';
 import { generateTestDataId } from 'utils/helpers';
 import { TestId } from 'utils/types';
 
-import { Props } from '../../Chip.types';
+import { ChipProps } from '../../Chip.types';
```
kostasdano, line 5: same

kostasdano (approved):
Code is LGTM but please check the GitHub Action Errors

src/components/Chip/components/Badge/Badge.tsx
```diff
@@ -2,14 +2,14 @@ import React from 'react';
...
 import { badgeStyle } from './Badge.style';
```
whosyourtaco, line 7: Isn't this type already exported from somewhere else in ictinus? Is this necessary?

whosyourtaco (approved):
LGTM. Check comments

src/components/Chip/components/Badge/Badge.tsx
```diff
@@ -2,14 +2,14 @@ import React from 'react';
...
 
 type TestProps = {
   dataTestId?: TestId;
 };
```
whosyourtaco, line 11: Isn't this type already exported somewhere else in the repo? Is this necesarry?

panvourtsis: yes, but this is not in the scope of the PR

panvourtsis:
as i wrote on the channel tests are meant to fail because there are several types that was renamed and are part of other PR's I will fix in the last PR

PR #545, opened 2022-12-02
[NDS-495] feat: change the second portion of components

panvourtsis:
## Description

This is part TWO of the goal, to fix all internal types so they can be easily accessible on the root of the library by an external user.


Christos-Zacharopoulos (approved):
There is a bunch of stuff breaking but since this is the second part of this chain of PR's it will probably be resolved in later PR's as discussed.

PR #546, opened 2022-12-02
[NDS-495] feat: change the third portion of components

panvourtsis:
## Description

This is part THREE of the goal, to fix all internal types so they can be easily accessible on the root of the library by an external user.


kostasdano (approved):
Code is LGTM but please check the GitHub Action Errors

panvourtsis:
as I wrote on the channel tests are meant to fail because there are several types that were renamed and are part of other PR's I will fix in the last PR

PR #547, opened 2022-12-02
[NDS-495] feat: change the 4th portion of components

panvourtsis:
## Description

This is part FOUR of the goal, to fix all internal types so they can be easily accessible at the root of the library by an external user.

### DIFF
--- a/src/components/Chart/BarChart/components/CustomTooltip/CustomTooltip.tsx
+++ b/src/components/Chart/BarChart/components/CustomTooltip/CustomTooltip.tsx
@@ -2,12 +2,12 @@ import React, { useState, useRef, useCallback, useEffect } from 'react';
 
 import { tickStyle, tooltipStyle, tooltipArrowStyle } from './CustomTooltip.style';
 
-type Props = {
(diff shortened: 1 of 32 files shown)


whosyourtaco (approved):
LGTM. Only issue here are the failed checks from github actions

src/components/Chart/BarChart/components/CustomTooltip/CustomTooltip.tsx
```diff
@@ -2,12 +2,12 @@ import React, { useState, useRef, useCallback, useEffect } from 'react';
 
 import { tickStyle, tooltipStyle, tooltipArrowStyle } from './CustomTooltip.style';
 
-type Props = {
+type CustomTooltipProps = {
```
whosyourtaco, line 5: Same comment https://github.com/Orfium/orfium-ictinus/pull/548/files#r1038470424

panvourtsis:
as i wrote on the channel tests are meant to fail because there are several types that was renamed and are part of other PR's I will fix in the last PR

PR #548, opened 2022-12-02
[NDS-495] feat: change the 5th portion of components

panvourtsis:
## Description

This is part FIFTH of the goal, to fix all internal types so they can be easily accessible at the root of the library by an external user.

Based on the ticket, what you are expecting to see is:

- Change all default Props naming we use on each component to ComponentNameProps so we can export and import easily
- Export all types and exported functions by default from each component
- Make all types by default exportable

PR #550, opened 2022-12-02
[NDS-495] feat: change the 7th portion of components

panvourtsis:
## Description

This is the SEVENTH part and last one of the goal, to fix all internal types so they can be easily accessible at the root of the library by an external user.

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -43,7 +43,7 @@ export type Selection = string | number;
 
 export type TableType = 'normal' | 'nested-header';
 
-type Props<T> = {
+type TableProps<T> = {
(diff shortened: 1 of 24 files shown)


src/components/Table/Table.tsx
```diff
@@ -43,7 +43,7 @@ export type Selection = string | number;
 
 export type TableType = 'normal' | 'nested-header';
 
-type Props<T> = {
+type TableProps<T> = {
```
whosyourtaco, line 46: Same comment https://github.com/Orfium/orfium-ictinus/pull/548/files#r1038470424

whosyourtaco (approved):

PR #552, opened 2022-12-02
[NDS-491 & NDS-492] theme reconfiguration typography

kostasdano:
## Description

[NDS-491](https://orfium.atlassian.net/browse/NDS-491) 
[NDS-492](https://orfium.atlassian.net/browse/NDS-492) 

[part II & III of [NDS-489](https://orfium.atlassian.net/browse/NDS-489)]

In this PR the **new typography configuration** is introduced and all typography usages are changed.
That includes: 
- `font-size` (and all usages)
- `font-weight` (and all usages)
- `font-family`
- `line-height`
- `letter-spacing`
- `text-case` (text-transform) (and all usages)
- `text-decoration`

This PR also includes a refactor to the `getFigmaTokensValue` in order to include all possible cases (for more info check the code comments)

### DIFF
--- a/src/theme/utils.ts
+++ b/src/theme/utils.ts
@@ -115,12 +115,9 @@ export const getFigmaTokensValue: {
     switch (type) {
       case FigmaTokenValueType.Pixels:
         return rem(Number(get(figmaTokensObject, [val, 'value'], '0'))) as string;
-        break;
       case FigmaTokenValueType.String:
         return get(figmaTokensObject, [val, 'value'], '0') as string;
-        break;
       case FigmaTokenValueType.Number:
         return Number(get(figmaTokensObject, [val, 'value'], '0')) as number;
-        break;
     }
   };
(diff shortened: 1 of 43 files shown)


src/theme/utils.ts
```diff
@@ -77,11 +77,25 @@ export const getColorErrors = [
...
 
 /**
- *
  * @param figmaTokensObject The parsed objects from Figma Tokens in the src/theme/constants/ dir
- * @returns the value of the figma token item converted into rem
+ * @param type 'pixels' refers to all the string values that need to be converted to rem
+ *             'string' refers to all the string values that need no conversion (e.g. opacity, font-family etc)
+ *             'number' refers to all the string values that need to be converted to numbers (e.g. font-weight)
  */
 export const getFigmaTokensValue =
```
panvourtsis, line 85: this is a great addition, i will update the type so we can avoid all the code to double define what is returned in every use

panvourtsis (approved):

PR #553, opened 2022-12-06
[NDS-495] new exportation of types

panvourtsis:
## Description

This is the main branch of new exportation Parts mention below **to show that tests are correct before merging**
The review was done separately on the below branches

The goal is to fix all internal types so they can be easily accessible on the root of the library by an external user.

#544 #545 #546 #547 #548 #549 #550


kostasdano (approved):
👌

PR #565, opened 2023-01-03
[NDS-518]: import new colors

kostasdano:
## Description

[NDS-518](https://orfium.atlassian.net/browse/NDS-518)

- Import new colors (lightPurple, neutral, gradient), update current colors with new bases and generate all new shades.
- Add comments for future obsolete code (will be removed when all components are revisited for v5)
- Fix types


src/theme/palette.ts
```diff
@@ -80,27 +84,30 @@ export const mainTypes = [
...
-    secondary: generatedColorShades;
-    light: generatedColorShades;
+    primary: GeneratedColorShades;
+    secondary: GeneratedColorShades;
+    light: GeneratedColorShades;
   };
-  flat: flatPalette;
-  pale: palePalette;
-
+  pale: PalePalette;
```
panvourtsis, line 107: also a TODO here for removal

panvourtsis (approved):

src/theme/palette.ts
kostasdano: the comment on line 101 refers to all the lines until line 110

PR #567, opened 2023-01-11
Feat/nds 493 bundling

panvourtsis:
## Description
This PR is solving the issue of performance bundling on our library. We fixed all the exportation of the library and in this step we export only the usable part of it. Because of this we decreased the size of the library by 1mb.

Also we tried to make it tree-shakeable. By adding tree shake we can now allow our apps to use separate components only and split them to different bundles as well without making them load everything at first load.

#### Before:
Media engagement was loading the whole library on the main bundle and all other bundles where loading from the main. This  had impact on the first load and on the download as well of the package.
Screenshot 2023-01-10 at 9 18 21 PM

#### After:
We can see here that the size was decreased and that separate ictinus components where moved to different bundles.
Screenshot 2023-01-10 at 9 11 48 PM

-------
### Size
Also size is significant smaller by exporting only of what is going to be used

<table>
<tr>
	<td>before</td>
	<td>after</td>
<tr>
	<td>
Screenshot 2023-01-11 at 4 19 47 PM
</td>
	<td>
Screenshot 2023-01-11 at 4 19 52 PM
</td>
<tr>
	<td> [master version](https://www.npmjs.com/package/@orfium/ictinus) </td>
	<td> [next version](https://www.npmjs.com/package/@orfium/ictinus/v/5.0.0-next.10) </td>
</table>

## Test Plan
There is no test plan to share, only the bundle analyzer could be something else but there is none for now


src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -2,15 +2,16 @@ import { ThemeProvider as EmotionThemeProvider } from '@emotion/react';
 import { css, Global } from '@emotion/react';
 import emotionReset from 'emotion-reset';
 import { ThemeSwitchProvider, useThemeSwitch } from 'hooks/useThemeSwitch';
-import { keys, merge, pick } from 'lodash';
+import keys from 'lodash/keys';
+import merge from 'lodash/merge';
+import pick from 'lodash/pick';
```
mkarajohn, line 7: 99.99% sure this is unnecessary and `import { keys, merge, pick } from 'lodash';` already treeshakes

panvourtsis: i run the analyzer and it was exporting all lodash before this

mkarajohn: TIL lodash exports in commonjs.
in that case maybe we can use this https://www.npmjs.com/package/@optimize-lodash/rollup-plugin

panvourtsis: nice one i will check it out, seems that it solves the issue

kostasdano (approved):
🎉

PR #570, opened 2023-01-23
[NDS-532] Button Tokens and new variants

kostasdano:
## Description

[NDS-532](https://orfium.atlassian.net/browse/NDS-532) 

[Figma Designs](https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/Design-System-V5.0---Working-File?node-id=574%3A7239&t=ttwvinRcqpbHAiPv-1)

**This PR introduces the new `Button Tokens` and the new `Text Button` with all its states and variants.**

- Introducing `Button Tokens`.
- Introducing `States Tokens`: States like `inactive`, `active`, `hover` have explicit styles that shall be used in all components. New buttons use these new tokens, but we keep the previous implementation for now. When other components are revisited (for the v5 updates) we shall replace the old state usages with the new tokens.
- Revisit Button and ButtonBase and update the props (remove obsolete ones).
- Introduce the new Button variants (types).
- Small fix in tokens parser for colours: could not parse neutral or gradient colours until now.
- Introduce new Loading Button (update the existing ButtonLoader)
- Update all Button usages.
- Update Snapshots.

## Screenshots

Screenshot_160

Screenshot_161

Screenshot_162

buttonsLoading

[NDS-532]: https://orfium.atlassian.net/browse/NDS-532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Button/Button.style.ts
```diff
@@ -1,36 +1,10 @@
...
-
-import { ButtonProps } from './Button';
+import { rem } from 'theme/utils';
 
 export const buttonSpanStyle = () => () => {
   return {
     display: 'flex',
     alignItems: 'center',
     justifyContent: 'center',
+    gap: rem(8),
```
panvourtsis, line 8: why we dont use global spacing (4)?

src/components/Button/Button.tokens.ts
```diff
@@ -0,0 +1,73 @@
...
+    wrapper: globals.borderRadius.get('0'),
+    icon: globals.borderRadius.get('5'),
+  },
+  borderWidth: {
+    1: globals.borderWidth.get('1'),
+    2: globals.borderWidth.get('2'),
+  },
+  color: {
+    primary: {
+      borderColor: semantic.borderColor.get('decorative.transparent'),
```
panvourtsis, line 46: all `semantic` references like `semantic.textColor` should be `theme.tokens.textColor` otherwise in any theme change this will be semantic

src/components/Button/ButtonLoader/ButtonLoader.style.ts
```diff
@@ -1,9 +1,62 @@
...
+      to: {
+        left: '100%',
+      },
+    });
+
+    return css({
+      width: '50%',
+      background: 'linear-gradient(270deg, #F814E1 0.02%, #4945EE 100%)',
+      height: getBarHeight({ type }),
+      boxShadow: `0 -2px 0px 0px inset ${stateTokens.active.backgroundColor[type]}`,
```
panvourtsis, line 56: this is common in all please use one reference at the top in a constant

src/components/ButtonBase/ButtonBase.style.ts
```diff
@@ -1,51 +1,51 @@
...
-
-    const backGroundColor = '#123456';
-    const borderWidth = 0;
+export const buttonWrapperStyle = ({
+  isBlock,
+}: Pick<ButtonBaseProps, 'isBlock'>): SerializedStyles =>
+  css({ position: 'relative', width: isBlock ? '100%' : 'fit-content' });
 
+export const buttonBaseStyle =
+  ({
```
panvourtsis, line 13: i expect that we miss the `circular` functionality here that the icon button needs

src/components/ButtonBase/ButtonBase.style.ts
```diff
@@ -1,51 +1,51 @@
...
+      padding: `${buttonTokens.spacing.textButton.paddingVertical} ${buttonTokens.spacing.textButton.paddingHorizontal}`,
+      borderRadius: buttonTokens.borderRadius.text,
+      border: `solid ${buttonTokens.borderWidth[1]} ${buttonTokens.color[type].borderColor}`,
       cursor: 'pointer',
       transition: 'background-color,border 150ms linear',
-      ':focus-visible:not(:disabled)': {
-        outline: getFocus({ theme, borderWidth: borderWidth }).styleOutline,
-      },
+
+      ':focus-visible:not(:disabled)': stateTokens.focus,
```
panvourtsis, line 31: also here you miss the `getFocus` functionality where you passed the `theme` in order to pick `semantic` as the theme from `theme.tokens`

src/components/Button/Button.tsx
```diff
@@ -1,42 +1,46 @@
...
+import Icon, { AcceptedIconNames } from 'components/Icon';
 
-export type onClickProp = { onClick: ClickHandler };
-export type ButtonProps = Omit<ButtonBaseProps, 'onClick'> &
+export type ButtonProps = ButtonBaseProps &
   TestProps &
-  onClickProp &
-  CommonButtonProps;
+  CommonButtonProps & {
+    /** An optional icon to put on the right of the button */
```
panvourtsis, line 14: why we moved this from `ButtonBaseProps` to here

src/components/ButtonBase/ButtonBase.tsx
```diff
@@ -49,33 +49,33 @@ const ButtonBase = React.forwardRef<HTMLButtonElement, ButtonBaseProps>((props,
...
-      onBlur={onBlur}
-      disabled={isDisabled || isLoading}
-    >
-      {children}
-    </button>
+    <div css={buttonWrapperStyle({ isBlock })}>
+      {isLoading && !isDisabled && <ButtonLoader type={type} dataTestId={testIdName} />}
+      <button
+        ref={ref}
+        type={buttonType}
```
panvourtsis, line 59: we changed the type to `buttonType` for consistency with HTML attributes so this could work -> https://www.w3schools.com/tags/att_button_type.asp

src/components/IconButton/IconButton.tsx
```diff
@@ -7,7 +7,7 @@ import Icon from '../Icon';
 import { AcceptedIconNames } from '../Icon/types';
 import ButtonBase, { ButtonBaseProps } from 'components/ButtonBase/ButtonBase';
 
-export type IconButtonProps = Omit<ButtonBaseProps, 'isIconButton' | 'iconLeft' | 'iconRight'> & {
+export type IconButtonProps = Omit<ButtonBaseProps, 'iconLeftName' | 'iconRightName'> & {
```
panvourtsis, line 10: Icon doesnt work as expected 
Screenshot 2023-01-27 at 2 27 46 PM

src/components/Notification/NotificationVisual/NotificationVisual.tsx
```diff
@@ -41,10 +41,14 @@ const NotificationVisual: React.FC<NotificationVisualProps> = ({
           css={actionContainer()}
           data-testid={generateTestDataId('visual-secondary', dataTestId)}
         >
-          <Button onClick={secondaryCTA}>{secondaryCTALabel}</Button>
+          <Button type="tertiary" onClick={secondaryCTA}>
```
panvourtsis, line 44: the type prop here is wrong based on HTML semantic

src/theme/states/states.tokens.ts
```diff
@@ -0,0 +1,70 @@
...
+    backgroundColor: Record<ComponentVariant, string>;
+  };
+  hover: {
+    backgroundColor: Record<ComponentVariant, string>;
+  };
+};
+
+const tokens: StatesTokens = {
+  focus: {
+    outline: `${semantic.borderColor.get('interactive.focused')} auto ${globals.borderWidth.get(
```
panvourtsis, line 30: i saw that this was used in other files but as stated before you have to pass the `theme` here and pick "semantic" from `theme.tokens`. This way it will use the **active** theme for getting the value if it is **semantic** or **dark** etc

panvourtsis (changes requested):
really nice work, some important things like button type attribute and how you pick theme variables and we are golden

src/components/Button/Button.tokens.ts
kostasdano: nice point 🚀 I will change all instances

src/components/ButtonBase/ButtonBase.style.ts
kostasdano: this will be added in a different, dedicated for the IconButton PR

src/components/Button/Button.tsx
kostasdano: it seemed to me better to move the leftIcon, rightIcon and avatar props to the Button component since:
1. they are no longer rendered inside the ButtonBase (so there is no point picking them from the ButtonBase for the types declaration)
2. semantically speaking: these props are only included in the Button variation of the ButtonBase; ButtonBase should be more abstract and act more as a wrapper for Button, IconButton or any future Button-like component. Since IconButton declares and renders its icon inside the IconButton, I thought it would be more clear to declare Button's leftIcon, rightIcon and avatar props (and components) inside Button too, alongside its other properties (e.g.: children). 

thoughts on this? 🤔

src/components/IconButton/IconButton.tsx
kostasdano: IconButton will be implemented in a dedicated future PR

src/theme/states/states.tokens.ts
kostasdano: yeap, you are totally right, I will change all instances of semantic across the project

panvourtsis (approved):
LGTM 🎊

PR #580, opened 2023-02-23
fix: date picker style

tsevdos:
## Description

The current DatePicker component (https://ictinus.herokuapp.com/?path=/story/design-system-datepicker--day-picker-with-predefined-values) has a max-width of 280px. By removing it we can control the with of the element on "application" level (it's containing div / parent).


tsevdos:
It was decided not to proceed with this implementation since it's a breaking change for the current apps.

tsevdos:
Revisiting of this may occur on the next version

PR #585, opened 2023-03-14
[NDS-594] Multiselect [v4.5]

kostasdano:
## Description

[NDS-594](https://orfium.atlassian.net/browse/NDS-594)

In this PR the MultiSelect with simple and async search is implemented:

- The MultiselectTextField is introduced: a TextField clone which supports multiple chips and text as input
- MultiselectTextField usage inside Select.
- Storybook update.
- Simple unit test.
- Missing eslint rule addition.

## Screenshots
Screenshot_35

[NDS-594]: https://orfium.atlassian.net/browse/NDS-594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -106,7 +106,7 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps & TestProps
     const inputRef = React.useRef<HTMLInputElement>(null);
     const combinedRefs = useCombinedRefs(inputRef, ref);
 
-    const initialValue = defaultValue || selectedOption;
+    const initialValue = defaultValue ?? selectedOption;
 
     const [inputValue, setInputValue] = React.useState(initialValue);
     const [searchValue, setSearchValue] = React.useState('');
@@ -171,6 +171,14 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps & TestProps
 
     const handleOnInput = React.useCallback(
       (event: ChangeEvent) => {
+        /** 
+         * For Multiselect: [for now] when we select an option the SelectMenu closes but the user
+         * can still type on the input field (so they must be able to see the SelectMenu)
+         */
+        if (!open) {
+          setOpen(true);
+        }
+
         handleSearch({
           event,
           isSearchable,
@@ -180,7 +188,7 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps & TestProps
           minCharactersToSearch,
         });
       },
-      [debouncedOnChange, isAsync, isSearchable, minCharactersToSearch]
+      [debouncedOnChange, isAsync, isSearchable, minCharactersToSearch, open]
     );
 
     const filteredOptions = useMemo(() => {
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -64,6 +64,7 @@ export type Props = {
   /** If true the user can select multiple options */
   multi?: boolean;
   /** The selected options in case of multiSelect */
+  /** @TODO merge selectedOption with selectedOptions in v5 */
   selectedOptions?: SelectOption[];
 } & TextFieldProps;
 
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -105,8 +105,13 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps & TestProps
     const [open, setOpen] = React.useState(false);
     const inputRef = React.useRef<HTMLInputElement>(null);
     const combinedRefs = useCombinedRefs(inputRef, ref);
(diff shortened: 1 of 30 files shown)


src/components/Select/components/MultiselectTextField/MultiselectTextField.tsx
```diff
@@ -0,0 +1,204 @@
...
+      if (!label && placeholder) {
+        if (required) {
+          return `${placeholder} *`;
+        }
+
+        return placeholder;
+      }
+
+      return label;
+    }, [label, placeholder, required]);
```
tsevdos, line 79:
```suggestion
    const inputPlaceholder = useMemo(() => {
      if (!label && placeholder) {
           return required ? `${placeholder} *` :  placeholder;
      }

      return label;
    }, [label, placeholder, required]);
```

src/components/Select/components/MultiselectTextField/MultiselectTextField.tsx
```diff
@@ -0,0 +1,204 @@
...
+      [disabled, locked, onOptionDelete, selectedOptions]
+    );
+
+    const icon = useMemo(() => {
+      if (rightIcon) {
+        if (typeof rightIcon === 'string') {
+          return <Icon name={rightIcon} size={20} color={theme.utils.getColor('lightGrey', 650)} />;
+        }
+
+        return rightIcon;
```
tsevdos, line 131:
```suggestion
       return typeof rightIcon === 'string' ? <Icon name={rightIcon} size={20} color={theme.utils.getColor('lightGrey', 650)} /> : rightIcon;
```

src/components/Select/components/MultiselectTextField/MultiselectTextField.tsx
```diff
@@ -0,0 +1,204 @@
...
+    const iconName = useMemo(() => {
+      if (locked) {
+        return 'lock';
+      }
+
+      if (hasValue) {
+        return 'close';
+      }
+
+      return 'search';
```
tsevdos, line 90: opinion:

```suggestion
   return locked ? 'lock' : hasValue ? 'close' : 'search';
```

src/components/Select/Select.tsx
```diff
@@ -97,15 +105,46 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps & TestProps
...
     const inputRef = React.useRef<HTMLInputElement>(null);
     const combinedRefs = useCombinedRefs(inputRef, ref);
-    const [inputValue, setInputValue] = React.useState(defaultValue || selectedOption);
+
+    const initialValue = defaultValue || selectedOption;
+
+    const [inputValue, setInputValue] = React.useState(initialValue);
     const [searchValue, setSearchValue] = React.useState('');
 
+    const textFieldValue = searchValue || inputValue.label;
```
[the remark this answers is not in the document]
kostasdano: Since `searchValue` and `inputValue.label` are both strings, if `searchValue === ''` (which is true on the initialization of searchValue) and let's say `inputValue.label === something with length >0`, then `searchValue ?? inputValue.label` would return `searchValue`, but `searchValue || inputValue.label` would return `inputValue.label` which is the correct one

src/components/Select/components/MultiselectTextField/MultiselectTextField.tsx
kostasdano: there is another return statement below so the ternary would not be the best option here, as it would increase the complexity of the code

src/components/Select/components/MultiselectTextField/MultiselectTextField.tsx
kostasdano: sonarqube doesn't like nested ternaries 😛

tsevdos (approved):

src/components/Select/Select.tsx
```diff
@@ -57,8 +57,14 @@ export type Props = {
...
   isVirtualized?: boolean;
-  /** A callback that's called when the user clicks the 'clear' icon */
+  /** A callback that's called when the user clicks the 'clear' icon of the (Single) Select, or the 'clear all' button of the MultiSelect */
   onClear?: () => void;
+  /** A callback that's called when the user clicks the 'clear' icon of a specific Chip in MultiSelect */
+  onOptionDelete?: (option: SelectOption) => void;
+  /** If true the user can select multiple options */
+  multi?: boolean;
+  /** The selected options in case of multiSelect */
+  selectedOptions?: SelectOption[];
```
panvourtsis, line 67: Can we add a @TODO here for v5 to merge selectedOptions with selectedOption? Value maybe []

PR #589, opened 2023-03-15
[NDS-613] make DatePicker a controlled component [v4.5]

tsevdos:
[NDS-613](https://orfium.atlassian.net/browse/NDS-613)

## Description

On the current DatePicker component, if you select a date (or range), you cannot reset it's value (`{ from: undefined, to: undefined }`). You can only change it ti another value. This creates problems in scenarios that you want to reset a form and you have already select a value into the DatePicker component. This PR resolves the above issue.

## Screenshot

chrome-capture-2023-2-15

## Test Plan

You must have a "controlled DatePicker" hence the value of the `DatePicker` must be controlled from it's parent component. Then simply reset it's value to ` { from: undefined, to: undefined }`. Everything else must work as expected. 

[NDS-613]: https://orfium.atlassian.net/browse/NDS-613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/DatePicker/DatePicker.stories.mdx
+++ b/src/components/DatePicker/DatePicker.stories.mdx
@@ -256,6 +256,16 @@ This datepicker has a predefined initial value that can be used both for initial
   </Story>
 </Preview>
 
+# Reset DatePicker
+
+<Preview>
+  <Story name="Reset DatePicker">
+    <Stack vertical>
+      <DatePickerShowcase />
+    </Stack>
+  </Story>
+</Preview>
+
 # Dynamic disableDates with all options available and custom label - Knobs
 
 <Preview>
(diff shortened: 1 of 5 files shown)


src/components/DatePicker/DatePicker.stories.mdx
```diff
@@ -270,3 +271,13 @@ This datepicker has a predefined initial value that can be used both for initial
     />
   </Story>
 </Preview>
+
+# Reset DatePicker (from parect component)
```
panvourtsis, line 275: please move this before knobs story, there is an unwritten rule for stories that we have. The last story is the story with all the variables(knobs)

panvourtsis (approved):

src/components/DatePicker/DatePicker.stories.mdx
tsevdos: fixed 8b5b3b2b59991b3d8c2db406a34ea21576814f01

kostasdano (approved):
🚀

PR #592, opened 2023-03-21
[NDS-626 - Select] keep SelectMenu open when toggling options [v4.5]

kostasdano:
## Description

[NDS-626](https://orfium.atlassian.net/browse/NDS-626)

## Screenshots

select

[NDS-626]: https://orfium.atlassian.net/browse/NDS-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -124,7 +124,6 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps & TestProps
     } = useMultiselectUtils({
       selectedOptions,
       options,
-      setOpen,
       setSearchValue,
       isSearchable,
       onClear,
@@ -144,9 +143,8 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps & TestProps
         );
       } else {
         setInputValue(option);
+        setOpen(false);
       }
-
-      setOpen(false);
 
       if (isSearchable) {
         setSearchValue('');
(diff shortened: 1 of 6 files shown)


src/components/Select/Select.tsx
```diff
@@ -124,7 +124,6 @@ const Select = React.forwardRef<HTMLInputElement, Props & InputProps & TestProps
     } = useMultiselectUtils({
       selectedOptions,
       options,
-      setOpen,
```
panvourtsis, line 127: do you need to remove this and keep your state? Can we have this logic on the hook?

src/components/utils/PositionInScreen/hooks.ts
```diff
@@ -17,18 +17,40 @@ export const usePositionInScreen = (
...
+   * We use this ResizeObserver in order to track any changes on the parent's height:
+   * This is necessary for the case of the MultiSelect, where the height of the TextField is dynamic
+   * and will increase/decrease as more Chips (Selected Options) are added/deleted. 
+   * Therefore the parentHeight is stored on the useState above.
+   */
+  useEffect(() => {
+    if (!parentRef.current) return;
+
+    const resizeObserver = new ResizeObserver((entries) => {
+      setParentHeight(entries[0].contentRect.height);
```
panvourtsis, line 32: `entries[0]` may fail. Can you use `head` ?

panvourtsis (changes requested):

src/components/Select/Select.tsx
kostasdano: this state is needed in the Select component (so it cannot stay only in the scope of the hook) and besides that, the `useMultiselectUtils` hook holds all the necessary logic for the MultiSelect field. the `[open, setOpen]` state applies to all Select field types so semantically speaking it should be on the Select component level

panvourtsis (approved):

PR #596, opened 2023-03-27
[NDS-644] fix: select component dropdown misposition [v4.5]

### LINKED ISSUE #595: Select Dropdown List Misposition
**_What?_**
When the SelectMenu is positioned above the TextField and the user searches for a specific term that makes the SelectMenu to have less height than its max-height, the SelectMenu is mispositioned.

multiselect-bug

kostasdano:
## Description

[NDS-644](https://orfium.atlassian.net/browse/NDS-644)

Resolves #595


panvourtsis (approved):
LGTM

PR #597, opened 2023-03-29
[FE-58] feat: add useSearchQueryParams hook

tsevdos:
## Description

The `useSearchQueryParams` hook, returns the URL search query parameters as a key - value plain object (**Record<string, string>**). Keep in mind that this hook can only be used into a `react-router-dom` context.

## Screenshot

localhost_6006__path=_docs_hooks-usesearchqueryparams--use-search-query-params

## Test Plan

- Unit testing


panvourtsis (approved):
LGTM 🎊

PR #603, opened 2023-04-10
[NDS-659] fix: select resize observer loop limit exceeded [v4.5]

kostasdano:
## Description

[NDS-659](https://orfium.atlassian.net/browse/NDS-659)

[NDS-659]: https://orfium.atlassian.net/browse/NDS-659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/utils/PositionInScreen/hooks.ts
+++ b/src/components/utils/PositionInScreen/hooks.ts
@@ -1,6 +1,8 @@
 import head from 'lodash/head';
 import { useEffect, useState } from 'react';
 
+import { resizeObserverHandler } from './utils';
+
 const useHeights = (
   /** Ref of the parent element */
   parentRef: React.MutableRefObject<any>,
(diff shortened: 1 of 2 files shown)


src/components/utils/PositionInScreen/hooks.ts
```diff
@@ -45,6 +52,10 @@ const useHeights = (
     const childResizeObserver = new ResizeObserver((entries) => {
       const dropdown = head(entries);
 
+      if (dropdown?.target === dropdown?.contentRect) {
```
panvourtsis, line 55: this function and :25 are very similar, I would suggest changing them to one to maintain one code with a callback action for your sideEffect

```
resizeObserverHandler = (callback) => new ResizeObserver((entries) => {
      const parent = head(entries);

      /** If the event is triggered by the element itself then return
       * to avoid an infinite loop
       */
      if (parent?.target === parent?.contentRect) {
        return;
      }

      if (parent) {
        callback(parent.contentRect.height);
      }
    });
```

and use of it

```
const childResizeObserver = resizeObserverHandler(setChildHeight);
...

```

PR #604, opened 2023-04-12
[NDS-660] Update button with comp tokens [v5]

kostasdano:
## Description

[NDS-660](https://orfium.atlassian.net/browse/NDS-660)


src/components/Button/Button.stories.mdx
```diff
@@ -111,7 +112,13 @@ Buttons with icons. 3 different variations based on the icon's placement.
   <Story name="Buttons with Icons" parameters={{ decorators: [withKnobs] }}>
     <Stack>
       <Button
-        iconLeftName={'dashboard'}
+        iconLeftName={select(
```
panvourtsis, line 115: can we use this as a function an just pass `iconLeftName` or `iconRightName`

src/theme/states/states.tokens.ts
panvourtsis, line 1: the `https://ictinus-pr-604.dev.orfium.xyz/?path=/story/guide-global-states--page` needs to change based on this now

panvourtsis (changes requested):

panvourtsis (approved):

PR #606, opened 2023-04-20
[NDS-665] feat: create MultiSelectBase component [v4.5]

kostasdano:
## Description

[NDS-665](https://orfium.atlassian.net/browse/NDS-665)

In this PR the MultiSelectBase component is introduced and it is rendered inside the existing MultiSelect component.

Practically its the MultiSelectTextField renamed to MultiSelectBase with a few changes and additions:
- the `isInteractive` prop declares whether the component will change UI-wise when in a different state (hover, focus etc).
- the component utils are moved inside a custom hook to reduce component size.

[NDS-665]: https://orfium.atlassian.net/browse/NDS-665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/MultiSelectBase/MultiSelectBase.tsx
```diff
@@ -57,40 +59,25 @@ const MultiselectTextField = React.forwardRef<HTMLInputElement, Props & InputPro
...
-      if (value === '' && event.key === 'Backspace') {
-        onOptionDelete();
-      }
-    };
+    const { inputPlaceholder, handleKeyDown, icon, hasLabel, TextfieldRef } =
+      useMultiSelectBaseUtils({
+        label,
+        placeholder,
+        required,
+        locked,
```
panvourtsis, line 74: for some reason based on our rules I expect this to be `isLocked`

Also I suggest hooks are made for multi filter use? Because here is used only once, therefore makes it obsolete

panvourtsis (changes requested):

panvourtsis (approved):

PR #608, opened 2023-04-27
[NDS-668] create multi filter component [v4.5]

kostasdano:
## Description

[NDS-668](https://orfium.atlassian.net/browse/NDS-668)

- Seperate the multi and single filter logic by creating 2 components: SingleFilter (contains the current functionality of the Filter) and MultiFilter (adds all the new functionality for the multi Filter)
- Render those 2 components conditionally inside the Filter component.
- Showcases
- Tests

[NDS-668]: https://orfium.atlassian.net/browse/NDS-668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kostasdano:
> More of a DQA comment but the paddings inside each selected value chip (?) seem to be a bit bigger than on the designs

The chips used in the current designs are based on v5. For this implementation we use an altered version of the v4.5 chips. This has been discussed with the DS team.

src/components/Filter/hooks/useMultiFilterUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+  onClear,
+  setSearchValue,
+  onFilterDelete = () => {},
+}: Props) => {
+  const [availableMultiFilters, setAvailableMultiFilters] = useState<FilterOption[]>([]);
+
+  const [multiFilters, setMultiFilters] = useState(selectedItems);
+
+  useEffect(() => {
+    if (multi) {
```
whosyourtaco, line 33: I get the way this hook is used, but by naming it `useMultiFilterUtils` you'd expect that whenever this hook runs, `multi` is `true`. Maybe this needs to be renamed, as it is used in the Filter.tsx and not in the MultiFilter component, or change the way it's used? Thoughts?

src/components/Filter/hooks/useMultiFilterUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+      } else {
+        setFilterLabel(`${items[0].label} + ${items.length - 1} more`);
+      }
+    } else {
+      setFilterLabel(defaultValue.label);
+    }
+  };
+
+  const handleOptionDelete = (option?: FilterOption) => {
+    let newItems: FilterOption[] = [];
```
whosyourtaco, line 55: The general consensus on using `let` is, and correct me if I'm wrong, that you should only use it when you know that a variable will change. In this case, the variable is a constant. I'd advocate for this to be turned into two `const` variables declared in each if statement, or another approach altogether.

whosyourtaco: Also consider wrapping this function in useCallback

src/components/Filter/types.ts
```diff
@@ -42,6 +47,11 @@ export type Props = {
   isLoading?: boolean;
   /** Whether the Options List is Virtualized or not */
   isVirtualized?: boolean;
+ /** If true the user can select multiple filter */
```
whosyourtaco, line 50: typo 'filters'

whosyourtaco (approved):
Nice job 🚀

src/components/Filter/components/SingleFilter/SingleFilter.style.ts
```diff
@@ -0,0 +1,19 @@
...
+    top: ${rem(48)};
+    min-width: ${rem(280)};
+    left: 0;
+    height: auto;
+    border-radius: ${theme.spacing.xsm};
+    background-color: ${theme.palette.white};
+    box-shadow: ${theme.elevation['02']};
+    z-index: 500;
+    overflow: hidden;
+    min-width: 100%;
```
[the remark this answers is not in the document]
kostasdano: nice catch

src/components/Filter/hooks/useMultiFilterUtils.ts
kostasdano: this hook is created to contain all the necessary callbacks and states for the case where `multi=true`. Therefore it's named `useMultiFilterUtils`. We need these functions and states on the Filter level (and not MultiFilter) so it cannot be called anywhere else. It was created to reduce the size of the `Filter` component. It follows the pattern of `useMultiSelectUtils` in Select component

PR #609, opened 2023-04-27
Feat/typography additions

panvourtsis:
## Description

This PR introduce some additions to the typography component based on Figma designs such as 
* rename `role` to `variant` for better control and documentation of component
* better prop documentation
* introduce `forwardRef`
* easy definition for italic, bold and underline
* introduce `type` that represent the text color based on that
* introduce `isInverted` for represent the inverted version of colors https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/DS-%7C-V5.0-%5BWIP%5D?node-id=16189-79638&t=DEGyBsm2FinPZemP-0

https://www.figma.com/file/8kMPBNYHHXz2AtkzeeDmk5/DS-%7C-V5.0-%5BWIP%5D?node-id=14861-61126&t=DEGyBsm2FinPZemP-0

## Test Plan

Snapshots

### DIFF
--- a/src/components/Typography/Typography.style.ts
+++ b/src/components/Typography/Typography.style.ts
@@ -1,10 +1,26 @@
 import { css, SerializedStyles } from '@emotion/react';
 import { Theme } from 'theme';
 
-import { TypographyType } from './Typography';
+import { TypographyVariant, TextColorTypes } from './Typography';
+import textColorFigma from '../../theme/tokens/semantic/variables/textColor';
+import { DotKeys } from '../../theme/tokens/utils';
 
 export const typographyWrapper =
-  ({ type }: { type: TypographyType }) =>
+  ({
+    variant,
+    isInverted,
+    isItalic,
+    isBold,
+    isUnderline,
+    type,
+  }: {
+    variant: TypographyVariant;
+    isInverted?: boolean;
+    isItalic?: boolean;
+    isBold?: boolean;
+    isUnderline?: boolean;
+    type: TextColorTypes;
+  }) =>
   (theme: Theme): SerializedStyles => {
     // headlines
     const headline01 = css`
@@ -13,6 +29,8 @@ export const typographyWrapper =
       font-size: ${theme.globals.typography.fontSize.get('9')};
       font-weight: ${theme.globals.typography.fontWeight.get('bold')};
       letter-spacing: ${theme.globals.typography.letterSpacing.get('0')};
+      margin-block-start: 1rem;
+      margin-block-end: 1rem;
     `;
     const headline02 = css`
       font-family: ${theme.globals.typography.fontFamily.get('roboto')};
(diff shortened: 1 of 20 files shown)


src/components/Typography/Typography.style.ts
```diff
@@ -126,5 +152,14 @@ export const typographyWrapper =
...
     };
 
-    return allStyles[type];
+    const textColorCategory = isInverted ? 'inverted' : 'light';
+    const textColor = `${textColorCategory}.${type}` as DotKeys<typeof textColorFigma>;
+
+    return css`
+      ${allStyles[variant]};
+      font-style: ${isItalic ? 'italic' : undefined};
+      font-weight: ${isBold ? 'bold' : undefined};
```
kostasdano, line 161: should this be `bold` or some value from our global font-weight ones?

src/components/Typography/Typography.style.ts
```diff
@@ -126,5 +152,14 @@ export const typographyWrapper =
...
+      text-decoration: ${isUnderline ? 'underline' : undefined};
```
kostasdano, line 162: _[optional]_ same here. we have textDecoration tokens for that. pretty much the same, but we can do it for consistency reasons.

kostasdano (approved):

src/components/Typography/Typography.style.ts
panvourtsis: this will implement `bold` or will not be defined at all, thus if there are global CSS will be applied

src/components/Typography/Typography.style.ts
panvourtsis: fixed here [`03c7ba6` (#609)](https://github.com/Orfium/orfium-ictinus/pull/609/commits/03c7ba6f8d3f4a3eb76fdc6e2c401766f55c50c6)

src/components/Typography/Typography.style.ts
```diff
@@ -26,113 +26,103 @@ export const typographyWrapper =
     const headline01 = css`
       font-family: ${theme.globals.typography.fontFamily.get('roboto')};
       line-height: ${theme.globals.typography.lineHeight.get('10')};
-      font-size: ${theme.globals.typography.fontSize.get('9')};
+      font-size: ${theme.globals.typography.fontSize.get('10')};
       font-weight: ${theme.globals.typography.fontWeight.get('bold')};
-      letter-spacing: ${theme.globals.typography.letterSpacing.get('0')};
-      margin-block-start: 1rem;
-      margin-block-end: 1rem;
+      letter-spacing: ${theme.globals.typography.letterSpacing.get('0')}px;
```
kostasdano, line 31: based on the `theme.globals.typography.letterSpacing.get` implementation, letterSpacing values from tokens are expected to be in px and then converted to rem. So `theme.globals.typography.letterSpacing.get('0')` should return rem, without the need of a px suffix.

By checking `src/theme/globals/constants/letterSpacing.ts` I see that the values are in a wrong and not parsable format (e.g. letterSpacing[2] = .25), that may have something to do with the fact that `px` suffix is needed here. 

We should revisit the letterSpacing tokens and how they are imported from Figma and parsed by the codebase in order to not need suffixes with px or rem.

Thoughts on this?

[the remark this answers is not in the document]
panvourtsis: fixed here [`eaf6e11` (#609)](https://github.com/Orfium/orfium-ictinus/pull/609/commits/eaf6e11ba7adf485271ac7d8a513c8096d51c37c)

PR #617, opened 2023-05-16
[NDS-670]: fix: fixes issue where onSort was called on initial render

whosyourtaco:
## Description

- Fixes an issue where the `onSort` callback was always called once during the initial rendering phase of the component, which wasn't expected or ideal.
- General cleanup of the file and completed a TODO item

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -1,7 +1,7 @@
 import useBreakpoints from 'hooks/useBreakpoints';
 import head from 'lodash/head';
 import pluralize from 'pluralize';
-import React, { useEffect, useState } from 'react';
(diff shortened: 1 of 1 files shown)


src/components/Table/Table.tsx
```diff
@@ -173,19 +166,25 @@ function Table<T>({
...
       if (prevState.column !== column) {
+        onSort?.(column, 'asc');
+
         return {
           column,
           order: 'asc',
         };
       }
 
+      onSort?.(column, prevState.order === 'asc' ? 'desc' : 'asc');
```
panvourtsis, line 186: `onSort` is called twice? 

1) Will be called on both line 169 or 178 
2) line 186

Why is that?

PR #618, opened 2023-05-16
[NDS-685]: docs: adds a .yalc section in the docs for active package development

whosyourtaco:
## Description

- Adds a section in DS's documentation that details the steps needed in order to develop the package using `yalc`

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "styled-components": "^5.2.0",
     "ts-jest": "^26.5.6",
     "ts-loader": "^9.1.2",
+    "tsc-watch": "^5.0.3",
     "typedoc": "^0.22.12",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -84,6 +84,7 @@
     "styled-components": "^5.2.0",
     "ts-jest": "^26.5.6",
     "ts-loader": "^9.1.2",
```
panvourtsis, line 86: missing yalc dependency as well

whosyourtaco: We are installing it as a global dependency on our system, per the instructions in the documentation. Do we really need to include it as a dependency inside the project? 

Since we use `yarn global add yalc`, we can just use `yalc` in every project on our system.

panvourtsis (approved):

PR #622, opened 2023-05-19
multi TextField suggestions

tsevdos:
## Description

* minor CSS fix
* correctly update multi TextField selected values 

This PR introduces a new Prop to handle multiple values named `multiValuesHandler`. This `multiValuesHandler` by default is being passed the value but based on user needs it will change this to whatever handling a user might want. 

In the case of the showcase provided, we change a paste of text like `text1, text2` to multiple values on `enter`

Also it fixes the `sx` issue for multiValue case

### DIFF
--- a/src/components/MultiTextFieldBase/MultiTextFieldBase.tsx
+++ b/src/components/MultiTextFieldBase/MultiTextFieldBase.tsx
@@ -1,4 +1,5 @@
 import useTheme from 'hooks/useTheme';
+import { merge } from 'lodash';
 import omit from 'lodash/omit';
 import React, { useMemo } from 'react';
 import { generateUniqueID } from 'utils/helpers';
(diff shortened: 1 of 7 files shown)


whosyourtaco (commented):
Nice feature! Works great.

On a side note, I've noticed two things with the component. First of all, if you just press `Enter`, new empty options will be created which I guess shouldn't be allowed. 

The other thing is that if there are multiple chips(?) with the same text typed in them, clearing one clears them all. We probably don't have a use case where we want to select the same options, but maybe it indicates that the implementation of the `Clear option` feature has a possible bug somewhere?

As these are not really related with the PR here, I'm just commenting them.

@panvourtsis @kostasdano  Should we create a ticket for these? Are they relevant? Or should we fix them on this PR?

panvourtsis:
I suggest moving forward with the suggestion and creating those as separate fixes on our component. Good catch @whosyourtaco!

panvourtsis (approved):

PR #623, opened 2023-05-23
[NDS-157] feat: custom scrollbar [v4.5]

kostasdano:
## Description

[NDS-157](https://orfium.atlassian.net/browse/NDS-157)

[NDS-157]: https://orfium.atlassian.net/browse/NDS-157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/ThemeProvider/ThemeProvider.style.ts
+++ b/src/components/ThemeProvider/ThemeProvider.style.ts
@@ -1,0 +1,51 @@
+import { SerializedStyles, css } from '@emotion/react';
+import { normalize } from 'polished';
+import { Theme } from 'theme';
+
+export const scrollbar = (theme: Theme): SerializedStyles => css`
+  // for Firefox
+  * {
+    scrollbar-width: thin;
+  }
+
+  // for Chrome
+  ::-webkit-scrollbar {
+    width: ${theme.spacing.sm};
+    height: ${theme.spacing.sm};
+  }
+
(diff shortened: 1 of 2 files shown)


src/components/ThemeProvider/ThemeProvider.style.ts
```diff
@@ -0,0 +1,51 @@
+import { SerializedStyles, css } from '@emotion/react';
+import { normalize } from 'polished';
+import { Theme } from 'theme';
+
+export const scrollbar = (theme: Theme): SerializedStyles => css`
+  // for Firefox
+  * {
+    scrollbar-width: thin;
+  }
```
whosyourtaco, line 9: This property is applied to **every** element in the DOM. Are we sure we need this?

panvourtsis: @kostasdano

panvourtsis (approved):

src/components/ThemeProvider/ThemeProvider.style.ts
kostasdano: We need this since we need ALL scrollbars to inherit this style; on top of that we tested this version on products and everything runs smoothly so this doesn't create any issues

PR #627, opened 2023-05-26
Update main.yml

panvourtsis:
## Description

fixing github deprecation set-output and save-state while removing old caching mechanism

[github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
[github.com/marketplace/actions/yarn-install-cache](https://github.com/marketplace/actions/yarn-install-cache)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,20 +29,10 @@ jobs:
       - name: Checkout 📥
         uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


whosyourtaco (approved):
👍

PR #629, opened 2023-06-01
[NDS-657] Feat/box

panvourtsis:
## Description

Introduce new helper component named `Box`. This component will take CSS props inline to help quickly prototype containers with no styled components.

## Test Plan
Storyshots


src/components/Box/Box.types.ts
```diff
@@ -0,0 +1,338 @@
...
+  | 'space-between'
+  | 'space-around'
+  | 'space-evenly'
+  | 'stretch'
+  | 'safe center'
+  | 'unsafe center'
+  | 'inherit'
+  | 'initial'
+  | 'unset';
+type AlignItems =
```
kostasdano, line 26: [optional] is there a way maybe to group types which equal the same unions?

e.g.:
type AlignBase = 'normal' | 'stretch| ........

type AlignItems = AlignBase
type AlignSelf = AlignBase

thoughts?

panvourtsis: are not 100% the same and even if they are is a really small win imo

PR #643, opened 2023-07-03
[NDS-706] feat: Textarea [v5]

kostasdano:
## Description

[NDS-706](https://orfium.atlassian.net/browse/NDS-706)

[NDS-706]: https://orfium.atlassian.net/browse/NDS-706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/TextArea/TextArea.tsx
```diff
@@ -28,40 +34,70 @@ export type TextAreaProps = {
...
+    </div>
+  ) : undefined;
 
   return (
     <React.Fragment>
       <TextInputBase {...props} sx={sx}>
-        <div css={{ width: '100% ' }}>
+        <div css={{ width: '100%' }}>
           <textarea
+            readOnly={isLocked || isReadOnly}
```
panvourtsis, line 77: use `role="textbox"` and i think we should use `aria-multiline` as true because we don't have single like only

src/components/TextArea/TextArea.tsx
```diff
@@ -28,40 +34,70 @@ export type TextAreaProps = {
...
-} & Pick<TextInputBaseProps, 'status'> &
+} & Pick<HTMLTextAreaElement, 'cols' | 'rows'> &
+  Pick<TextInputBaseProps, 'status' | 'label'> &
   TestProps;
 
 const TextArea = React.forwardRef<HTMLTextAreaElement, TextAreaProps>((props, ref) => {
   const {
     id = undefined,
-    placeholder = '',
+    label,
```
panvourtsis, line 44: i think we need a typed way to get inputbase props and others so we know if something is missing

src/components/TextArea/TextArea.tsx
```diff
@@ -28,40 +34,70 @@ export type TextAreaProps = {
...
+    placeholder,
     isRequired = false,
+    isReadOnly,
     isDisabled,
+    status,
     isResizeEnabled = true,
```
panvourtsis, line 50: is this the default behavior? are we sure?

src/components/TextArea/TextArea.tsx
```diff
@@ -28,40 +34,70 @@ export type TextAreaProps = {
   onKeyDown?: React.KeyboardEventHandler<HTMLTextAreaElement | HTMLInputElement>;
   /** Callback fired when the `input` value typed is changed */
   onInput?: React.EventHandler<any>;
-} & Pick<TextInputBaseProps, 'status'> &
+} & Pick<HTMLTextAreaElement, 'cols' | 'rows'> &
```
panvourtsis, line 37: i dont like that we pick some of the input and some of the textarea props. As a user I might want to pass any textarea prop.

panvourtsis (changes requested):

src/components/TextArea/TextArea.tsx
kostasdano: after discussion with the Design Team, the implied that this is the default expected behavior.

panvourtsis (approved):

PR #669, opened 2023-09-14
[NDS-781] DatePicker & DateRangePicker Custom Keyboard Navigation Investigation [v5]

kostasdano:
## Description

[NDS-781](https://orfium.atlassian.net/browse/NDS-781)

This PR is the result of a spike investigation on custom keyboard navigation for DatePicker and DateRangePicker

- Refactor existing useKeyBoardEvents hook to include all arrow keyboard cases.
- Add Keyboard Navigation for DatePickInput and all necessary style changes for calendar button.
- Add Keyboard Navigation for the Calendar Grid so the user can navigate through all days.
- Add Keyboard Navigation for Month Selection (Button + List).

[NDS-781]: https://orfium.atlassian.net/browse/NDS-781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


panvourtsis:
i would expect that when i open the datepicker from the date icon, it would move the focus to either the first selected item or the current day

src/components/DatePicker/Month/Month.tsx
```diff
@@ -56,6 +57,65 @@ const Month: React.FC<MonthProps> = ({ year, month, onDaySelect, selectedDays, d
...
+
+  const { keyboardProps } = useKeyboardEvents({
+    events: {
+      keydown: {
+        onArrowUp: () => {
+          setFocusedDay((prevCursor) => {
+            if (prevCursor - 7 < 0) {
+              return prevCursor;
+            }
+
```
whosyourtaco, line 74: Part (or all) of the logic in this component could probably be moved to a separate hook to make this component easier to read, as the useKeyboardEvents is kinda hard to read into

whosyourtaco (approved):
LGTM

panvourtsis:
Screenshot 2023-09-22 at 11 44 33 AM
When i select in range two dates and I have selected with keyboard the last selected day is seems broken

kostasdano:
> Screenshot 2023-09-22 at 11 44 33 AM When i select in range two dates and I have selected with keyboard the last selected day is seems broken

after discussion with the Design, this case will stay as it is, UI-wise, just to indicate that the element is focused.

panvourtsis (approved):

PR #670, opened 2023-09-18
feat: upgrading dependencies of react to 18 

panvourtsis:
## Description

This PR tends to solve typing for React 18. Any app that uses ictinus till this point have issues when they are using our library. From now on we are solving this and we are also backward compatible with React 17

You should expect to see:
- a lot of changes of `React.FC` to our new type `ReactFCC` which is the new component with children. 
- many dummy removal of lines and emotion styled updated on snapshot files because of emotion upgrade

## Test Plan

Already tested with RC version on

- Rights Cloud
- Earnings


mkarajohn (approved):
Personally I think this would be a great opportunity to completely drop the pattern of "typed" React components; I would much prefer for them to be written as `function Component(props: Props) {}` but as long as the implemented solution resolves the reported issue, it's fine by me.

PR #674, opened 2023-09-26
[NDS-807] fix: select backspace does not clear value [v5]

kostasdano:
## Description

[NDS-807](https://orfium.atlassian.net/browse/NDS-807)

In this PR: 
- the Backspace issue is resolved: when Backspace is clicked on SingleSelect while the input value equals the selected option's label, then the Select will be cleared (that also fixes the issue of non-searchable Select where you cannot clear the field)
- as implemented on v4, the empty state of Select from now on is `undefined` ; so when the field is cleared, the onChange is called with `undefined` as an argument.
- Typos are fixed.
- Async Select Showcase wasn't working properly; now it's fixed.

[NDS-807]: https://orfium.atlassian.net/browse/NDS-807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -23,6 +23,8 @@ import PositionInScreen from 'components/utils/PositionInScreen';
 
 export const emptyValue: SelectOption = { label: '', value: '' };
 
+/** @TODO: Refactor component to reduce Cognitive Complexity */
+
 const Select = React.forwardRef<HTMLInputElement, SelectProps>((props, ref) => {
   const {
     selectedOption,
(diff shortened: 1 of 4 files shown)


src/components/Select/Select.tsx
```diff
@@ -81,6 +81,17 @@ const Select = React.forwardRef<HTMLInputElement, SelectProps>((props, ref) => {
         onBackspace: () => {
           debouncedOnChange('');
           setIsOpen(true);
+
+          /** When the textField's value equals the selectedOption's label we should clear the field*/
+          if (!isMulti) {
+            if (selectedOption && onChange) {
+              if (selectedOption.label === textFieldValue) {
```
panvourtsis, line 88: why 3 if here, just wondering?

kostasdano: mostly for readability reasons, although in the future this component will be refactored to reduce Cognitive Complexity

PR #677, opened 2023-10-02
[NDS-536] feat: Link Component

kostasdano:
## Description

[NDS-536](https://orfium.atlassian.net/browse/NDS-536)

- Introduce Link component.

[NDS-536]: https://orfium.atlassian.net/browse/NDS-536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Link/Link.tsx
+++ b/src/components/Link/Link.tsx
@@ -10,7 +10,7 @@ const Link = React.forwardRef<HTMLAnchorElement, LinkProps>((props, ref) => {
   const {
     htmlHref,
     type = 'primary',
-    placement = 'standalone',
+    placement = 'block',
     size = 1,
     iconName,
     isDisabled,
--- a/src/components/Link/Link.tsx
+++ b/src/components/Link/Link.tsx
@@ -1,0 +1,43 @@
+import useTheme from 'hooks/useTheme';
+import React from 'react';
+
+import { linkContainer } from './Link.style';
+import { LinkTokens, getLinkTokens } from './Link.tokens';
+import { LinkProps } from './Link.types';
+import Icon from 'components/Icon';
+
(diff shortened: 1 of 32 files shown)


panvourtsis:
Although we discussed it over the chat, I will make the notes here as well for future reference.

- Navigation with keyboard doesn't show which element is focusable.
- I am expecting to see the HTML anchor properties on the Link. Props like download, target etc.

src/components/Link/Link.tsx
```diff
@@ -0,0 +1,43 @@
+import useTheme from 'hooks/useTheme';
+import React from 'react';
+
+import { linkContainer } from './Link.style';
+import { LinkTokens, getLinkTokens } from './Link.tokens';
+import { LinkProps } from './Link.types';
+import Icon from 'components/Icon';
+
+const Link = React.forwardRef<HTMLAnchorElement, LinkProps>((props, ref) => {
```
panvourtsis, line 9: missing component extension for external 3rd party routers, ticket criteria quoted below
```
Link should be part of the Typography family and it could use its props maybe for further expansion
Link should be router agnostic. So removal of react-router-dom is necessary.
```

kostasdano: 1) Since (with the current designs and tokens) Link has very specific typography (label01-03), colors (primary, inverted) and html representation (anchor element plus whatever children the user provides) tbh I don't see any semantic or technical reason why Link and Typography (component) should be connected

2) react-router-dom is not used inside the Link component. as a package it is used by other ictinus components

panvourtsis (approved):
🚀

PR #680, opened 2023-10-06
[NDS-821] fix: deps update

panvourtsis:
## Description

Updating all the storybook packages to the latest of v6. All of our critical dependency alerts and some high ones so it increase our security


kostasdano (approved):
🚀

PR #682, opened 2023-10-09
[NDS-802] feat: tooltip component

kostasdano:
## Description

[NDS-802](https://orfium.atlassian.net/browse/NDS-802)


src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -18,72 +20,159 @@ import { FIGMA_URL } from '../../utils/common';
...
+                width: '300px',
+                height: '100px',
+                display: 'flex',
+                justifyContent: 'center',
+                alignItems: 'center',
+              }}
+            >
+              <Tooltip
+                content={'This is a Tooltip'}
+                placement={select('placement', ['top', 'right', 'bottom', 'left'], 'right')}
```
panvourtsis, line 125: in this story i would expect all 4 placements as showcase with no playground on this property

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -18,72 +20,159 @@ import { FIGMA_URL } from '../../utils/common';
...
+
+Text Tooltip takes a string as a content prop. There are 2 variations of Text Tooltips, default and inverted.
+
+<Preview>
+  <Story name="Text Tooltip">
+    <Stack>
+      <Function>
+        {() => {
+          return (
+            <div style={{ padding: '16px' }}>
```
panvourtsis, line 63: on keyboard navigation delay doesnt work as hover. It doesn't mind me at all its even better but it was worth mention it

src/components/Tooltip/Tooltip.stories.mdx
```diff
@@ -18,72 +20,159 @@ import { FIGMA_URL } from '../../utils/common';
   }}
 />
 
-# Tooltip
+<SectionHeader title={'Tooltip'} />
```
panvourtsis, line 23: i would expect also hover for chromatic screenshots 😄 like we did in select I think

src/components/Tooltip/Tooltip.tsx
```diff
@@ -2,59 +2,37 @@ import Tippy from '@tippyjs/react';
...
+      placement = 'right',
+      isInverted = false,
+      isInteractive = false,
+      delayIn = 500,
+      delayOut = 500,
+    },
+    ref
+  ) => {
+    return (
+      <Tippy
```
panvourtsis, line 21: test ids and rest?

kostasdano: test-ids: `<Tippy/>` component doesn't accept data-testid as a prop ([more info here](https://github.com/atomiks/tippyjs-react/issues/89)). In order to have test-id I guess the user would have to use the interactive tooltip and add test-id to the custom content.

rest: no need for rest prop, all tooltip props are extracted and used specifically inside the component.

panvourtsis (approved):

panvourtsis (approved):

PR #686, opened 2023-10-12
[NDS-804] feat: Tag component 

kostasdano:
## Description

[NDS-804](https://orfium.atlassian.net/browse/NDS-804)

[NDS-804]: https://orfium.atlassian.net/browse/NDS-804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Tag/Tag.stories.mdx
```diff
@@ -0,0 +1,223 @@
...
+The `isSelected` prop indicates whether the Tag is selected or not.
+
+<Preview>
+  <Story name="Selectable Tag">
+    <Stack>
+      <Function>
+        {() => {
+          const [isSelected, setIsSelected] = useState(false);
+          return (
+            <Tag isSelected={isSelected} onSelect={() => setIsSelected(true)}>
```
panvourtsis, line 109: I would change this to toggle, I would also provide the selected part from the initial state for both user and chromatic to capture

```suggestion
            <Tag isSelected={isSelected} onSelect={() => setIsSelected((isSelected) => !isSelected)}>
```

src/components/Tag/Tag.stories.mdx
```diff
@@ -0,0 +1,223 @@
...
+<Props of={Tag} />
+
+## Usage
+
+<UsageGuidelines
+  guidelines={[
+    'Use tags to show metadata that needs to be highlighted somehow',
+    'Use tags to relate secondary information about a page element (px a “Writer” tag next to a name)s',
+    'Use selectable tags to quickly filter existing content',
+    'Do not use tags as standalone elements in a page; they should always relate to something elseh',
```
panvourtsis, line 49: What i would expect for `do not use` cases is to have another section of being defined in another array here and display it accordingly. This applies to all sections that we have

For example
```
<UsageGuidelines guidelines={[]} policies={[]} //do not use cases />
```

src/components/Tag/hooks/useGetTagUtils.tsx
```diff
@@ -0,0 +1,87 @@
...
+    if (!isInteractive && iconName)
+      return <Icon size={12} name={iconName} color={tokens(`textColor.${color}` as const)} />;
+
+    return null;
+  }, [color, dataTestPrefixId, iconName, isInteractive, isSelectable, isSelected, tokens]);
+
+  const suffix = useMemo(
+    () =>
+      isClearable ? (
+        <div tabIndex={0} css={iconStyles()} onKeyDown={handleIconKeyDown}>
```
panvourtsis, line 64: I would change that to `tabIndex -1` to make it unfocusable because you are relying on the tag click not only the X. Therefore now when i navigate with my keyboard on tab it focuses on tag and on the second tag it doesnt go to the next element it focuses the X which has no focus style. 
```suggestion
        <div tabIndex={-1} css={iconStyles()} onKeyDown={handleIconKeyDown}>
```

src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    return (
+      <div
+        onClick={onSelect}
+        onKeyDown={handleKeyDown}
+        tabIndex={isInteractive ? 0 : -1}
+        css={tagContainerStyles({ size, color, isSelectable, isClearable, isSelected })}
+        ref={ref}
+        data-testid={`${dataTestPrefixId}_tag_container`}
+        {...rest}
```
panvourtsis, line 35: need for aria-label if none exist

panvourtsis (approved):

PR #690, opened 2023-10-24
[NDS-541] feat: Compact Buttons

kostasdano:
## Description

[NDS-541](https://orfium.atlassian.net/browse/NDS-541)

Add compact `Button`, `IconButton` and `DropdownButton`

[NDS-541]: https://orfium.atlassian.net/browse/NDS-541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Button/Button.tsx
```diff
@@ -39,13 +40,19 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>((props, ref) =>
...
     <ButtonBase {...props} ref={ref} isLoading={isLoading} onClick={onClick}>
       <span css={buttonSpanStyle()}>
-        {hasAvatar && <Avatar src={avatar?.src}>{avatar?.label}</Avatar>}
-        {iconLeftName && !hasAvatar && (
-          <Icon name={iconLeftName} color={tokens(`${type}.textColor` as ButtonTokens)} />
-        )}
-        <span>{children}</span>
-        {iconRightName && (
-          <Icon name={iconRightName} color={tokens(`${type}.textColor` as ButtonTokens)} />
+        {size === 'compact' ? (
```
panvourtsis, line 43: compact button doesnt have icon or avatar? this is really different.

I would expect documentation to reflect this as well as types

src/utils/stories.ts
```diff
@@ -6,8 +6,8 @@ import iconSelector from 'components/Icon/assets/iconSelector';
 export const getIconSelectorKnob = (propName: string, initialValue?: AcceptedIconNames) =>
   select(
     propName,
-    Object.keys(iconSelector)
+    Object.keys({ ...iconSelector, '': '' })
```
panvourtsis, line 9: can we give a `no icon` name to that because it took me awhile to understand it

src/components/Button/Button.tsx
kostasdano: according to figma docs `‘compact’ variant only supports simple text button (no avatars, icons), icon buttons, and dropdown buttons`.

I will add that to storybook also

src/utils/stories.ts
kostasdano: select know accepts an array of strings, so if we add `no icon` it will throw an error trying to find an icon named `no icon`. This is why empty string is added as an option (if empty string is passed as icon prop, no icon is rendered)

panvourtsis: please leave a comment on the code then

panvourtsis (approved):

PR #708, opened 2023-11-09
[NDS-867] Radio and RadioGroup

kostasdano:
## Description

[NDS-867](https://orfium.atlassian.net/browse/NDS-867)


src/__storyshots__/Updated Components/List/List-GroupList.snap
```diff
@@ -139,9 +139,9 @@
         style="height: 100%;"
       >
         <ul
-          aria-labelledby="react-aria-:rk:"
+          aria-labelledby="react-aria-:r24:"
```
panvourtsis, line 142: we need to revisit aria labels and ids

panvourtsis (changes requested):

panvourtsis (approved):

PR #722, opened 2023-11-28
feat/storybook-7-upgrade-11

panvourtsis:

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -1,6 +1,6 @@
-import { Elevation } from 'index';
+import type { Elevation } from 'index';
 import React from 'react';
-import { SpacingKey } from 'theme/globals/spacing';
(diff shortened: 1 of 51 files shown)


kostasdano (approved):

src/components/Card/Card.tsx
```diff
@@ -13,7 +13,12 @@ export type CardProps = {
   radius?: SpacingKey;
 };
 
-const Card: React.FCC<CardProps> = ({ elevated, isTransparent = false, radius, children }) => {
+const Card: React.FC<React.PropsWithChildren<CardProps>> = ({
```
kostasdano, line 16: this could probably be FCC

panvourtsis: changed here https://github.com/Orfium/orfium-ictinus/pull/722/commits/47a8ac4868aa9ef425efe5b4747e9db9d7656fc2 good catch

PR #735, opened 2023-11-29
fix: build storybook and prepublishonly script

panvourtsis:
## Description

This PR changes the way we build for Review Apps to not build twice by introducing `prepare` and `prepublishOnly`

Also it fixed an issue with `he.jsxDEV is not a function emotion` where we were building having `NODE_ENV=staging` for reviews


kostasdano (approved):
🎉

PR #755, opened 2024-01-11
[NDS-892] feat: Search

kostasdano:
## Description

[NDS-892](https://orfium.atlassian.net/browse/NDS-892)


src/components/Search/StatefulSearch.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { forwardRef, useState } from 'react';
+import isEqual from 'react-fast-compare';
+
+import Search from './Search';
+import type { SearchProps } from './Search.types';
+import type { FilterOption } from 'components/Filter';
+
+export const LABEL = 'Friends';
+export const options = [
```
panvourtsis, line 9: why are these being displayed only for the story showcasing?

kostasdano: check latest commit for fix

panvourtsis:
based on aria description i am expecting aria label for `searchbox` and some keyboard handling for clear with `Escape` key and `Enter` for submit

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/searchbox_role

https://react-spectrum.adobe.com/react-aria/useSearchField.html

panvourtsis (approved):

PR #759, opened 2024-01-26
feat(banner): add alternative cta / button

tsevdos:
## Description

[Ticket](https://orfium.atlassian.net/browse/NDS-901)

### DIFF
--- a/src/components/Notification/Banner/Banner.tsx
+++ b/src/components/Notification/Banner/Banner.tsx
@@ -19,6 +19,10 @@ export type Props = {
   closeCTA?: () => void;
   /** The data test id if needed */
   dataTestId?: TestId;
+  /** The alternative call-to-action label of the Notification */
+  alternativeCTALabel?: string;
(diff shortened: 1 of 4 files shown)


kostasdano (changes requested):
~`NotificationActions` type includes `secondaryCTALabel` and `secondaryCTA`.  You should use these instead of adding alternativeCTA. Check `Toast` or `Snackbar` Notifications for reference~

tsevdos:
Hello @kostasdano

I started the implementation as you state (with primary and secondary),  but after communicating with @panvourtsis we decided to go with the "alternative" name approach. Can you have a chat with him to be all on the same page. For me it is the exact same thing (minor change/fix).

best regards,
John

panvourtsis:
I  just have communicated with Kostas regarding this and he is now up to speed with the situation. Thanks!

kostasdano (approved):
at first I didn't understand the direction of the implementation but now we are at the same page, nice work!

PR #770, opened 2024-01-31
[NDS-900] docs: shaping up all props and styles of old and new

panvourtsis:


src/components/Modal/Modal.mdx
kostasdano, line 1: maybe we can consider moving ModalContent to a seperate page

src/components/Search/Search.types.ts
```diff
@@ -1,7 +1,12 @@
 import type { FilterProps, SingleFilterProps } from 'components/Filter';
 import type { TextFieldProps } from 'components/TextField';
 
-type SearchFilterConfig = {
+export type filterConfig = Pick<
```
kostasdano, line 4: is this type used somewhere? below there is a duplicate declaration (?)

kostasdano:
Charts component's stories should also be revisited to be aligned with the other stories styles (in terms of anchors on top of the page etc)

kostasdano (approved):

PR #787, opened 2024-03-11
[NDS-981] fix: select new option label error

kostasdano:
## Description

[NDS-981](https://orfium.atlassian.net/browse/NDS-981)

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -165,8 +165,18 @@ const Select = React.forwardRef<HTMLInputElement, SelectProps>((props, ref) => {
       }
     } else {
       if (onChange) {
-        // @ts-ignore
-        onChange(option);
+        if (option.isCreated) {
+          const newOption: SelectOption = {
+            value: option.value,
+            label: option.value.toString(),
+            isCreated: true,
+          };
+          // @ts-ignore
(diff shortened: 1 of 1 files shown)


src/components/Select/Select.tsx
```diff
@@ -165,8 +165,18 @@ const Select = React.forwardRef<HTMLInputElement, SelectProps>((props, ref) => {
...
       if (onChange) {
-        // @ts-ignore
-        onChange(option);
+        if (option.isCreated) {
+          const newOption: SelectOption = {
+            value: option.value,
+            label: option.value.toString(),
+            isCreated: true,
+          };
+          // @ts-ignore
```
panvourtsis, line 174: why we are ingoring here?

panvourtsis: @kostasdano reminder

kostasdano: its because of the conditional types used for onChange:
```
(single-select)
onChange?: (selectedOption?: SelectOption) => void;

(multi-select)
onChange?: (selectedOption: SelectOption[]) => void;
```

we had some issues with this approach so we left some ts-ignore comments needed in order for this implementation to work

panvourtsis (approved):

PR #807, opened 2024-04-30
[NDS-999] feat: table sticky header

kostasdano:
## Description

[NDS-999](https://orfium.atlassian.net/browse/NDS-999)

[NDS-999]: https://orfium.atlassian.net/browse/NDS-999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -58,9 +58,10 @@ const Table = <TData,>({
                   rowSize={rowSize}
                   width={header.getSize()}
                   {...(header.column.getCanSort() && {
-                    isSorted: header.column.getIsSorted().toString(),
+                    colSortingState: sorting?.sortingColumn?.find((col) => col.id === header.id),
                     onSort: header.column.toggleSorting,
                     isMultiSortable: header.column.getCanMultiSort(),
+                    resetSorting: header.column.clearSorting,
                   })}
                   sx={sx?.th}
                 >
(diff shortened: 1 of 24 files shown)


src/components/Table/Table.tsx
```diff
@@ -13,11 +13,24 @@ const Table = <TData,>({
...
 
   const hasColumnVisibilityConfig = Boolean(columnVisibility && setColumnVisibility);
 
+  /** If true, the scrollbar of tbody is visible */
+  const [hasScrollbar, setHasScrollbar] = React.useState(false);
+
+  const tBodyRef = useRef<HTMLTableSectionElement>();
+
+  React.useEffect(() => {
+    if (tBodyRef?.current) {
```
panvourtsis, line 29: do we need `useEffect` here? hasScrollBar could be true/false or null. If !null then it was set

kostasdano: we need it in order for the hasScrollbar to be initialised after table is rendered 🤔

panvourtsis: but you can use if without `useEffect` right?

kostasdano: if I dont put useEffect, I cannot get the values of scrollheight and clientHeight since tBodyRef is null on initial render. I use useEffect to trigger a re-render

panvourtsis (approved):

PR #811, opened 2024-05-17
[CDP-142] fix(codemods): small re-ordering for not false override

panvourtsis:
## Description

This PR fixes issues while I was migrating for testing purposes the earnings application. Issues fixed as follows

- Exportation from one file failed as imports were using paths that didn't exist
- Fixed package.json exportations to limit only to `@orfium/ictinus` and not to deep dive.
- Remove rollup because one file is not needed anymore
- codemods update for buttons to not override properties while migrating
- TextInputBase had `display: flex` but not `flex: 1` so the width was not full. **(Expect a lot of snapshot updates)**


kostasdano (approved):
🚀

PR #815, opened 2024-05-21
[NDS-994] feat: table pagination

kostasdano:
## Description

[NDS-994](https://orfium.atlassian.net/browse/NDS-994)

[NDS-994]: https://orfium.atlassian.net/browse/NDS-994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Table/components/TPagination/TPagination.style.ts
```diff
@@ -0,0 +1,60 @@
...
+  ({ isSticky }: Pick<TPaginationProps, 'isSticky'>) =>
+  (theme: Theme): SerializedStyles => {
+    return css`
+      display: flex;
+      align-items: center;
+      justify-content: space-between;
+      background: ${theme.tokens.colors.get('backgroundColor.default')};
+      height: ${rem(44)};
+      width: 100%;
+      padding: 8px 16px;
```
panvourtsis, line 19: 1) if this is fixed keep it in rem as pixel are not responsive
2) do we need tokens here?

src/components/Table/components/TPagination/TPagination.style.ts
```diff
@@ -0,0 +1,60 @@
...
+    align-items: center;
+    gap: 8px;
+  `;
+};
+
+export const counterContainer = (): SerializedStyles => {
+  return css`
+    display: flex;
+    align-items: center;
+    gap: 32px;
```
panvourtsis, line 45: rem and tokens?

src/components/Table/components/TPagination/TPagination.style.ts
```diff
@@ -0,0 +1,60 @@
...
+  `;
+};
+
+export const counterWrapper = (): SerializedStyles => {
+  return css`
+    display: flex;
+    gap: 4px;
```
panvourtsis, line 52: rem and tokens?

src/components/Table/types.ts
```diff
@@ -100,6 +104,29 @@ export type RowsConfig = {
...
+  /** OnChange callback for next/prev buttons */
+  onChange: (page: number) => void;
+  /** Whether go-to-first-page and go-to-last-page buttons are visible */
+  isEnhancedPaginationVisible?: boolean;
+  /** Manually disable next page */
+  isNextPageDisabled?: boolean;
+  /** Manually disable previous page */
+  isPrevPageDisabled?: boolean;
+  /** Show items per page list options */
+  itemsPerPageOptions?: SelectOption[];
```
panvourtsis, line 123: selectOption has many things

https://designlab-pr-815.orfium-dev.com/?path=/docs/updated-components-fields-select--overview

Do we need them all ?

src/components/Table/components/TTitle/TTitle.tsx
```diff
@@ -57,7 +57,7 @@ const TTitle: React.FCC<Props> = ({
         )}
       </div>
 
-      <div>
+      <div css={{ display: 'flex', gap: '16px' }}>
```
panvourtsis, line 60: keep in mind 'rem'

src/components/Table/components/TPagination/TPagination.style.ts
kostasdano: after merging master I will replace all these with tokens, I've put a TODO comment for now

src/components/Table/components/TPagination/TPagination.style.ts
kostasdano: same here

src/components/Table/components/TPagination/TPagination.style.ts
kostasdano: same here

src/components/Table/types.ts
kostasdano: since we are using Select component for itemsPerPage, shouldn't we use the necessary component types?

panvourtsis: we should but select comes with a lot of other properties. We can omit to secure the functionality as user doesn't have access to select of pagination

kostasdano: instead of `SelectOption` we could use `Omit<SelectOptionValues, 'iconProps'>`, 
which equals `{id: string | number, value: string}`. 

something like this maybe?

panvourtsis (approved):

PR #830, opened 2024-09-17
feat: inline alert v5

geokaralis:
## Description

`<InlineAlert />` v5.

## Accessibility

Adheres to the WAI-ARIA [alert](https://www.w3.org/WAI/ARIA/apg/patterns/alert/) pattern.


src/components/InlineAlert/InlineAlert.style.ts
```diff
@@ -0,0 +1,129 @@
+import { css } from '@emotion/react';
+import { rem } from 'polished';
+import type { Theme } from 'theme';
+
+import type { InlineAlertProps } from './InlineAlert.types';
+import { generateStylesFromTokens } from '../Typography/utils';
+
+const inlineAlertStyles = (props: InlineAlertProps) => (theme: Theme) => {
+  return css`
```
tsevdos, line 9:
```suggestion
const inlineAlertStyles = (props: InlineAlertProps) => (theme: Theme) => `
```

panvourtsis:
in terms of documentation i suggest 2 things
1) create a notification section and have inline and future notifications under it. This will be easier to have all in one place as per design as well.
2) Refer to the old notification that are deprecated and link to new section. Also code wise we can use `@deprecated` tag to notify us when we are about to use them

src/components/utils/useDOMRef.ts
```diff
@@ -0,0 +1,16 @@
+import { useImperativeHandle, useRef } from 'react';
+import type { RefObject } from 'react';
+
+export function createDOMRef<T extends HTMLElement = HTMLElement>(ref: RefObject<T | null>): T {
```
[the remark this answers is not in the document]
geokaralis: Thanks for also providing the `useCombinedRefs`. `useDOMRef` operates slightly different, using only a single ref and making it accessible to both the parent and child. Added more info on the JSDoc comment.

.eslintignore
```diff
@@ -13,4 +13,4 @@ __mocks__
 ./vite.config.ts
 ./vite.codemods.config.ts
 ./rollup.config.js
-!.storybook
+.storybook
```
geokaralis, line 16: Temporarily disable eslint on `.storybook`.

src/components/InlineAlert/InlineAlert.tsx
```diff
@@ -0,0 +1,78 @@
...
+      onDismiss,
+      testId,
+      hasAutoFocus,
+      ...otherProps
+    } = props;
+    const theme = useTheme();
+    const dismissId = useId();
+    const domRef = useDOMRef(ref);
+
+    const actionElements =
```
geokaralis, line 26: `useMemo` was unnecessary here, as the parent will always rerender the actions.

src/components/InlineAlert/InlineAlert.mdx
```diff
@@ -0,0 +1,45 @@
...
+    'To provide contextual information to the user pertaining to their current task.',
+    'In conjunction with form validation - When you need to learn about a feature or information that is relevant to your current activity.',
+    'When you need assistance in fixing errors, often in conjunction with form validation.',
+  ]}
+  policies={[
+    'If you need to deliver a global message from the admin to all users (e.g. maintenance update), use a broadcast message instead.',
+    `If you are immediately confirming a user's action while they remain in the same view, use a toast instead.`,
+    'If you are promoting, soliciting feedback, or informing the user about a feature, use a banner instead.',
+  ]}
+/>
```
kostasdano, line 45: 1. for documentation consistency, I suggest we put a new section here with all the variants, in order to have documentation per case (if needed). Check also other mdx files as guide for the structure of the page.

```
<SubsectionHeader title="Variants" />  /** Use this instead of hashtag-header for style purposes */

### Neutral

<Canvas of={InlineAlertStories.Neutral} />

....
```
2. We can merge `success, warning, error, informational, neutral` into one story named "Types" with documentation and rendering of all types in one story. This is an optional suggestion though.

eg:

```
<SubsectionHeader title="Variants" />  /** Use this instead of hashtag-header for style purposes */

### Types

There are 4 types of inline alert based on.........

<Canvas of={InlineAlertStories.Types} />

....
```

3. Really nice and interesting touch on adding controls table instead of `<ArgTypes of={InlineRadio} />`. This one adds our playground logic and the controls to the mdx level and not on a separate story as we already do in all the other components. This is a new documentation feature though and creates inconsistency. We have to either keep this and create ticket to change the documentation for all the other components in order to have the playground in the overview section of the mdx (as in the inline alert), or keep the current logic of having a separate Playground story and change it later for all components. @panvourtsis thoughts?

src/components/InlineAlert/InlineAlert.stories.tsx
```diff
@@ -0,0 +1,141 @@
...
+import Button from '../Button';
+import Link from '../Link';
+import { InlineAlert } from './InlineAlert';
+import { useState } from 'react';
+
+const Container = styled.div`
+  display: inline-flex;
+  flex-direction: column;
+  align-items: flex-start;
+  gap: 8px;
```
kostasdano, line 12: for gaps we should use the spacing tokens (values provided by design)

src/components/InlineAlert/InlineAlert.style.ts
```diff
@@ -0,0 +1,124 @@
...
+      ? theme.tokens.colors.get('backgroundColor.default')
+      : theme.tokens.colors.get('palette.secondary.muted')};
+    border-radius: ${theme.dimension.borderRadius.get('md')};
+    border: 1px solid ${theme.tokens.colors.get('borderColor.decorative.default')};
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
+    font-family: sans-serif;
+    min-height: var(--_min-height);
+    outline: none;
+    outline-offset: 2px;
```
kostasdano, line 34: if possible, you can use a token or at least rem function for this.

src/components/InlineAlert/InlineAlert.style.ts
```diff
@@ -0,0 +1,124 @@
...
+    transition: outline-offset 150ms cubic-bezier(0.4, 0, 0.2, 1);
```
kostasdano, line 35: I really like the cubic-bezier usage here. nice touch👌

src/components/InlineAlert/InlineAlert.tsx
```diff
@@ -0,0 +1,78 @@
...
+        {actionsArray.length > 0 ? <div css={styles.actions}>{actionsArray}</div> : null}
+        {onDismiss ? (
+          <Icon
+            role="button"
+            aria-label="Dismiss notification"
+            id={onDismiss ? dismissId : undefined}
+            name="close"
+            css={styles.dismiss}
+            onClick={onDismiss}
+            size={20}
```
kostasdano, line 70: is there a token for that? either comp or semantic?

src/components/InlineAlert/InlineAlert.types.ts
```diff
@@ -0,0 +1,14 @@
+import type { ReactElement, ReactNode } from 'react';
+
+export type AlertStatus = 'neutral' | 'informational' | 'error' | 'warning' | 'success';
+
+export interface InlineAlertProps {
+  status?: AlertStatus;
```
kostasdano, line 6: You can add a comment description for each comment (also renders on the ArgTable or ControlsTable in storybook and enriches the documentation)

kostasdano:
minor comments mostly for consistency from my side also, great work and and really nice touches introduced! welcome to DS 🎨 🎉

src/__storyshots__/Updated Components/List/List-VirtualizedList.snap
```diff
@@ -135,9 +135,9 @@
         style="height: 450px;"
       >
         <ul
-          aria-labelledby="react-aria-:r2n:"
+          aria-labelledby="react-aria-:r2v:"
```
kostasdano, line 138: are these IDs expected to change?

src/components/InlineAlert/InlineAlert.stories.tsx
geokaralis: I will remove the gap here, since it's not required as an example.

src/components/InlineAlert/InlineAlert.tsx
geokaralis: I followed the documentation from here `Interactive icon's size is a prop provided by the user; suggested sizes (as shown below) are 12 (xs), 16 (sm), 20 (m), 24 (l)`. Ideally we could add support for fixed sizes `sm`, `md` for the `InteractiveIcon` component at some point.

As for the `data-testid`, since we have specified aria attributes for the dismiss button, are you ok doing the following:
```js
const { container } = render(<InlineAlert onDismiss={onDismiss}>With dismiss</InlineAlert>);

const closeButton = within(container).getByRole('button', { name: 'Dismiss notification' });
```
It would be more preferred not to add a `dismissTestId` prop in this case, what do you think?

src/__storyshots__/Updated Components/List/List-VirtualizedList.snap
geokaralis: This is a good question, I also noticed that some snapshots changed irrelevant of the button and inline alert. Could it be that these were stale and hadn't be updated? I also use the `useId()` hook to bind the dismiss button on the alert. I think react aria uses the same. Maybe this is also a potential reason.

src/components/InlineAlert/InlineAlert.tsx
kostasdano: you are right about the Icon size, I forgot the Interactive Icon's api tbh 😁 

about the test-id my only concern is that this might be confusing for the QA since they use data-testids regularly.

src/components/InlineAlert/InlineAlert.mdx
geokaralis: Adjusted documentation to include variants and `ArgTypes`

src/components/InlineAlert/InlineAlert.tsx
geokaralis: Included `dismissTestId` on [4428743](https://github.com/Orfium/orfium-ictinus/pull/830/commits/4428743037fdc10742380e9c08e611c6e83bc77c)

src/__storyshots__/Updated Components/List/List-VirtualizedList.snap
geokaralis: Added a custom serializer for deterministic react aria ids in snapshot tests on [4428743](https://github.com/Orfium/orfium-ictinus/pull/830/commits/4428743037fdc10742380e9c08e611c6e83bc77c)

src/components/InlineAlert/InlineAlert.tsx
kostasdano: we have a `TestProps` type in `utils/types` that we add to all components for consistency, you can add this

src/components/InlineAlert/InlineAlert.mdx
```diff
@@ -43,3 +44,25 @@ A message from the system to the user that offers contextual and timely informat
...
+
+<Canvas of={InlineAlertStories.Error} />
+
+### Warning
+
+<Canvas of={InlineAlertStories.Warning} />
+
+### Success
+
+<Canvas of={InlineAlertStories.Success} />
```
kostasdano, line 68: optionally you can add the playground as a story to be visible in the overview page

PR #836, opened 2024-10-07
feat: broadcast

geokaralis:
## Description

`<Broadcast />` v5.


src/components/Broadcast/Broadcast.style.ts
```diff
@@ -0,0 +1,137 @@
...
+    isolation: isolate;
+    padding: calc(${theme.dimension.spacing.get('md')} - var(--_border-width))
+      calc(${theme.dimension.spacing.get('md')} - var(--_border-width));
+    display: inline-grid;
+    grid-template-columns: auto minmax(var(--_min-content-width), 1fr) auto;
+    grid-template-areas: 'icon content actions close';
+    align-items: center;
+    background: ${theme.tokens.colors.get('palette.secondary.muted')};
+    border-radius: ${theme.dimension.borderRadius.get('md')};
+    border: 1px solid ${getBorderColor(props.status, theme)};
```
kostasdano, line 26: why `1px` and not `var(--_border-width)` ?

geokaralis: Good catch we missed this in `InlineAlert` as well. I will update it in both places.

kostasdano (approved):

PR #846, opened 2025-05-07
feat: update react-aria vite and eslint

geokaralis:
## Description

- Update `react-aria` to latest
- Update `vite` to match fusion
- Update `eslint` and resolve errors
- Update `node_version` in workflows


panvourtsis (approved):
only concern here is that we need to validate automation tests don't break because of data-testid removal

PR #865, opened 2025-06-12
fix(tooltip): set opacity to 1 for consistent visibility

panvourtsis:
## Description

this PR makes the opacity of tooltip 1 as the react-tooltip puts opacity by default


tsevdos (approved):

tsevdos:
"plakato"...

PR #901, opened 2025-09-08
chore: update tsconfig and eslint

geokaralis:
- Upgrade to Eslint 9
- `tsconfig.json` reference for ictinus lib in storybook


panvourtsis (approved):
very good

